PR #1, opened 2016-09-12
Tap 4 trust pinning

trishankkarthik:
Begin a PR to merge TAP 4 (Trust Pinning).


awwad:
https://github.com/theupdateframework/taps/blob/awwad-tap-0004-trust-pinning/tap4.md#pin-file

It's a subtle point, but the two comment lines here may be slightly misleading:
```
{
  "repositories": {
    "django": {
      // metadata would be at https://repository.djangoproject.com/metadata/
      // targets would be at  https://repository.djangoproject.com/targets/
      "url": "https://repository.djangoproject.com/",
```

Where the metadata and targets directories are expected to be is defined *on a mirror-to-mirror basis* in the repository_mirrors argument to the tuf.client.updater.Updater constructor, and so is flexible beyond this point. You could say "might be at" rather than "would be at". Three mirrors for the same (django, say) repository may have different directories for metadata or targets.

trishankkarthik:
On 14 September 2016 at 16:27, Sebastien Awwad <notifications@github.com>
wrote:

> https://github.com/theupdateframework/taps/blob/
> awwad-tap-0004-trust-pinning/tap4.md#pin-file
>
> It's a subtle point, but the two comment lines here may be slightly
> misleading:
>
> {
>   "repositories": {
>     "django": {
>       // metadata would be at https://repository.djangoproject.com/metadata/
>       // targets would be at  https://repository.djangoproject.com/targets/
>       "url": "https://repository.djangoproject.com/",
>
> Where the metadata and targets directories are expected to be is defined *on
> a mirror-to-mirror basis* in the repository_mirrors argument to the
> tuf.client.updater.Updater constructor, and so is flexible beyond this
> point. You could say "might be at" rather than "would be at". Three mirrors
> for the same (django, say) repository may have different directories for
> metadata or targets.
>

Ah, didn't know that. Please go ahead and edit it!

trishankkarthik:
Woops, accidentally closed it. Just go ahead and edit master :)

PR #7, opened 2016-10-25
LDeditsTAP3

jhdalek55:
OK. Here's my edited version. It's mostly word order changes that I think make the writing simpler and more direct.

I didn't incorporate any of Sebastien's  comments...figured that was better left to you.

### DIFF
--- a/tap3.md
+++ b/tap3.md
@@ -155,7 +155,7 @@ The "roles" attribute now specifies only the keys used by a role, and
 (optionally) the metadata filename corresponding to the role.
 If the filename is not specified, then it is assumed to be located at
 "<ROLENAME>.json".
-Different roles may share the same filename (e.g., "F1.json"), but will use a
+Different roles may share the same filename (e.g., "F1.json"), but may use a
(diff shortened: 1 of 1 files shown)


tap3.md
```diff
@@ -82,7 +81,7 @@ The following is the [previous format for a targets metadata file](https://githu
 
 ### Limitations of the previous format
 
-Note that, unfortunately, this format does not allow a list of targets to be
+The previous format prevents a list of targets from being
```
trishankkarthik: Thanks for the comment! Yes, @awwad had a similar concern earlier.

Would you please see if this has been improved in the [latest commit](https://github.com/theupdateframework/taps/blob/7d2557950df395dff87276535bc87efafa818ece/tap3.md)? If not, what would you recommend that we do?

vladimir-v-diaz: > This TAP allows targets to be delegated to a combination of roles on a repository, instead of just a single role.

This is much better.

trishankkarthik: Okay, cool, thanks!

PR #12, opened 2016-12-06
Update tap3.md

vladimir-v-diaz:
Revise the section on resolving multi-role delegations to mention what happens when the "terminating" attribute is True or False.

### DIFF
--- a/tap3.md
+++ b/tap3.md
@@ -219,24 +219,26 @@ targets:
 }
 </pre>
 
-## Resolving delegations
+## Resolving multi-role delegations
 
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
I noticed while working on the implementation that TAP 3 doesn't include the "terminating" attribute.  Apart from incorrectly excluding this required attribute, it might also be confusing to talk about the "terminating" attribute and not list it in the examples.

Assuming I'm not mistaken, should I go ahead and edit TAP 3 for include the missing "terminating" attribute?

JustinCappos:
Yes, it should be there.

Justin

PR #14, opened 2016-12-28
Update tap4.md

JustinCappos:
Starting to edit for clarity.

### DIFF
--- a/tap4.md
+++ b/tap4.md
@@ -13,11 +13,12 @@
 
 TAP 4 describes how users may specify that a certain repository should be used
 for some targets, while other repositories should be used for other targets.
-In other words, this TAP allows users to _map_ targets to repositories in a
-manner similar to how targets can be delegated to roles.
-This allows users to say that: (1) a target may be found on one of many
-repositories, each with a different root of trust, and / or (2) many
-repositories may be required to sign the target.
+In other words, this TAP allows users to _map_ target names to repositories in a
+manner similar to how targets with specific names can be delegated to different roles.
+This allows users to say that a target with a specific type of name (such as 
+django\* or \*.tar.gz) may be found on a specific repository.  Each repository
+has its own root of trust (root role, etc) so a compromise of one repository
+does not impact other repositories.
 
 # Motivation
 
--- a/tap4.md
+++ b/tap4.md
@@ -93,7 +93,8 @@ and is only modified by a user who is trusted to configure the TUF instance.
 The map file contains a dictionary that holds two keys, "repositories" and
 "mapping."
 
-The value of the "repositories" key is another dictionary.
+The value of the "repositories" key is another dictionary that lists the URLs
+for a set of repositories.
 Each key in this dictionary is a _repository name_, and its value is a list of
 URLs.
 The repository name also corresponds to the name of the local directory on the
@@ -108,7 +109,8 @@ repository shall be updated at all._
 These files would be updated following the steps detailed in
 [this section](#downloading-metadata-and-target-files).
 
-The value of the "mapping" key is a list.
+The value of the "mapping" key is a priority-ordered list that maps paths (i.e., 
+target names) to the specific repositories.
 Every member in this list is a dictionary with at least two keys:
 
 * "paths" specifies a list of target paths of patterns. A desired target must
@@ -132,6 +134,10 @@ The following is an example of a map file:
   // downloaded.
   "mapping": [
     {
+      // Much like target delegation, the order of these entries indicates
+      // the priority of the delegation.  The entries listed first will be
+      // considered first.
+      
       // Map any target matching *Django* to both Django and PyPI.
       "paths":        ["*django*"],
       "repositories": ["Django", "PyPI"],
@@ -165,7 +171,7 @@ the metadata file for a top-level role.
 It is up to the repository to enforce that every delegated targets role uses a
 unique name.
 
-All targets files would be stored under the "targets" directory.
+All targets files would be stored under the "targets" directory.   
 Beyond this, the repository may organize target files into any hierarchy it
 requires.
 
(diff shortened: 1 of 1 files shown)


tap4.md
```diff
@@ -103,9 +103,6 @@ would be found.
 The list of URLs specifies where TUF clients may download metadata and target
 files.
 Each URL points to a [directory containing metadata and target files](#metadata-and-targets-layout-on-repositories).
-Each URL may be a [file URI](https://en.wikipedia.org/wiki/File_URI_scheme),
-which means that these files shall be updated from a local directory on disk
-instead of a remote server.
```
trishankkarthik: Yes, file URIs would be in a later TAP.

tap4.md
```diff
@@ -13,21 +13,25 @@
...
 ## Use case 1: obtaining different targets from different repositories
 
-It is desirable to use the same instance of TUF to download and verify different
-targets hosted on different repositories (for example, Python packages from
-PyPI, and Ruby packages from RubyGems).
+Itmay be desirable to use the same instance of TUF to download and verify different
+targets hosted on different repositories (for example, some Python packages from
+their maintainers, while getting other Python packages from PyPI).  In this way,
+one can securely get all Python packages regardless of where they are hosted.
 There are significant advantages in using the same instance of TUF to manage
```
trishankkarthik: Sure. This tries to explain the use case you think is most important: reusing the same updater across multiple repositories. One can try to solve this problem using different updaters, one for each repository, which is the situation today, but that raises more problem than it solves. Does this make sense?

tap4.md
```diff
@@ -141,6 +147,9 @@ The following is an example of a map file:
       // the following mapping will be consulted.
     },
     {
+      // Some paths need not have a URL.  Then those paths will not be updated.
+      ...
+    {
```
trishankkarthik: Sorry, I don't understand. What is the example?

tap4.md
```diff
@@ -165,7 +171,7 @@ the metadata file for a top-level role.
 It is up to the repository to enforce that every delegated targets role uses a
 unique name.
 
-All targets files would be stored under the "targets" directory.
+All targets files would be stored under the "targets" directory.
```
trishankkarthik: This is strictly for the targets themselves, not the targets metadata. On second thought, though, we don't need this at all. The targets metadata is sufficient to say where the targets themselves are located.

tap4.md
```diff
@@ -223,11 +235,13 @@ on a repository.
...
 ensures that: (1) that it has been signed by a threshold of keys, and (2) it has
-not expired.
+not expired.   Recall that the URL field may either contain the location to 
+update the files, or may be empty to say that the repository metadata should not
+be updated.  We will now explicitly explain the procedure for doing this.
 Next, the client tries to update the root metadata file.
 Let M denote the list of URLs associated with this repository in the map file,
 and R denote the list of URLs associated with this top-level role (in this case,
-the root role) in the root metadata file.
+the root role) in the root metadata file.
```
trishankkarthik: I think there should be a single place that discusses exactly how metadata files are updated. Otherwise, we may run the risk of confusing readers, no?

tap4.md
```diff
@@ -223,7 +229,7 @@ on a repository.
 
 First, the client loads the latest downloaded [root metadata file](tap5.md), and
 ensures that: (1) that it has been signed by a threshold of keys, and (2) it has
-not expired.
+not expired.
```
trishankkarthik: Well, if it's expired, then I think what we do now is to try to update the root metadata file.

If it's inconsistent, meaning it's not signed by a threshold of keys it has specified for itself, then that is a strange denial-of-service attack. One should not trust this root metadata file at all, right? Even if we update the root metadata file, and find that the new root metadata file has been signed by a threshold of keys from the previous root metadata file, that still seems strange.

tap4.md
```diff
@@ -141,6 +150,9 @@ The following is an example of a map file:
       // the following mapping will be consulted.
     },
     {
+      // Some paths need not have a URL.  Then those paths will not be updated.
```
JustinCappos: I'm trying to say that you can omit the URL or repo to have paths not be updated, right?  At least that is what I understand from reading later.

trishankkarthik: Let me make sure I understand you. Do you mean that the list of URLs in both the root metadata file, and the map file, are both empty? Then, yes, no metadata file will be updated.

This raises an interesting question. In this case, where would targets themselves be fetched? From disk, or from the repository? I think it's fine either way, but the first way assumes that a TUF client always caches targets the same way it caches metadata.

tap4.md
```diff
@@ -13,21 +13,25 @@
...
-repositories, each with a different root of trust, and / or (2) many
-repositories may be required to sign the target.
+In other words, this TAP allows users to _map_ target names to repositories in a
+manner similar to how targets with specific names can be delegated to different roles.
+This allows users to say that a target with a specific type of name (such as 
+django\* or \*.tar.gz) may be found on a specific repository.  Each repository
+has its own root of trust (root role, etc.) so a compromise of one repository
+does not impact other repositories.  This TAP also discusses how the AND relation can
+be extended to multiple repositories to have multiple different repositories with 
+separate roots of trust need to sign off on the same target before installation.
```
ecordell: > This TAP also discusses how the AND relation can be extended to multiple repositories to have multiple different repositories with  separate roots of trust need to sign off on the same target before installation.

Sorry that this came so late after merging - but this sentence is a bit off. If "need to" were removed then it could make sense, but I also think it's weird that it references an "AND" relation as though it's a tuf-specific thing defined elsewhere. 

Perhaps:

> This TAP also discusses how multiple repositories with separate roots of trust can be required to sign off on the same target, effectively creating an AND relation.

PR #15, opened 2016-12-30
Move the download bits from TAP 4 to TAP 5.

trishankkarthik:
Remove restrictions of how files are kept on repositories and clients.

Remove restriction on the file names of delegated targets roles in the
snapshot metadata. This came up in a meeting where it was thought that
the top-level targets role, and all delegated targets roles, should be
listed as if they were in the same directory, in order to prevent
delegated targets roles from accidentally overwriting top-level roles.

Since these restrictions were lifted, backwards compatibility looks
quite good.

### DIFF
--- a/tap4.md
+++ b/tap4.md
@@ -1,7 +1,7 @@
 * TAP: 4
 * Title: The map file
 * Version: 1
-* Last-Modified: 14-Dec-2016
+* Last-Modified: 29-Dec-2016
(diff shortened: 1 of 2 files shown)


trishankkarthik:
@JustinCappos Made the changes you recommended. Haven't had time to think through everything, so let me know if I missed anything!

PR #16, opened 2017-01-01
TAP 4: The map file

JustinCappos:


tap4.md
```diff
@@ -194,16 +194,15 @@ the same targets metadata (i.e., length and hashes).
 
 # Backwards Compatibility
 
-This specification is backwards-compatible, because older clients need not be
-modified in order to recognize a map file.
+This specification is backwards-compatible, however older clients will not utilize the map file
+and so will ignore this TAP.
```
JustinCappos: Didn't the location of the repo's URL change in this TAP?  Wasn't it in a config file before this?

tap4.md
```diff
@@ -178,7 +178,7 @@ If all repositories in the current mapping have not signed any metadata
 about the target, then take one of the following two actions.
 If the "terminating" attribute is set to true, report that there is no metadata
 about the target.
-Otherwise, proceed to similarly interpret the next mapping.
+Otherwise, proceed to similarly interpret the next mapping.
```
JustinCappos: This whole paragraph is confusing to me.  Can you work with Lois on a writing pass here?

jhdalek55:
I'd be happy to take a look at it. I've been staying away from these
documents while the technical details were being developed. I can review
the whole thing this week if you'd like.

Lois

JustinCappos:
Okay, thanks!  It's mostly just a single paragraph.  I've been over the
rest and modified text as needed.

Justin

tap4.md
trishankkarthik: Okay, sent @jhdalek55 an email

tap4.md
trishankkarthik: Depends, I think.

If a client used the "low-level" updater, then I think the repo URLs are specified by the client as a [function parameter](https://github.com/theupdateframework/tuf/blob/7a18e0b9e996054970adbac94a37cb0d4faac314/examples/client/example_client.py#L41-L47).

If the client used the "high-level" interposition updater, then the same repo URLs are specified in a [JSON config file](https://github.com/theupdateframework/tuf/tree/develop/tuf/interposition#basic).

@vladimir-v-diaz Let me know if I missed something.

tap4.md
```diff
@@ -194,16 +194,15 @@ the same targets metadata (i.e., length and hashes).
...
+and so will ignore this TAP.
 Older clients may continue to use a single repository.
 New clients need to add relatively little code to interpret the map file.
 New clients simply need to be careful to store the metadata for each repository
 separately from other repositories (e.g., by using a different directory for
 each repository).
 
-An existing repository needs not change how it already stores metadata and
-targets using the TUF specification prior to this TAP.
+An repository need not change in any way to support this TAP.
```
trishankkarthik: "A~~n~~ repository..."

tap4.md
trishankkarthik: By the way, the example says:

> // Some paths need not have a URL.  Then those paths will not be updated.

Are we still using this example? Is the list of repo URLs allowed to be empty?

trishankkarthik:
(Copying comment from @ecordell)

> This TAP also discusses how the AND relation can be extended to multiple repositories to have multiple different repositories with  separate roots of trust need to sign off on the same target before installation.

Sorry that this came so late after merging - but this sentence is a bit off. If "need to" were removed then it could make sense, but I also think it's weird that it references an "AND" relation as though it's a tuf-specific thing defined elsewhere. 

Perhaps:

> This TAP also discusses how multiple repositories with separate roots of trust can be required to sign off on the same target, effectively creating an AND relation.

tap4.md
trishankkarthik: This is what we came up with:

# Searching for targets using the map file

In order to search for targets on repositories using the map file, a TUF client should perform the following steps:
1. Look at the first mapping in the list of mappings.
2. If the desired target matches a targets path pattern in the "paths" attribute, then download and verify metadata from every repository specified in the "repositories" attribute.
3. Ensure that the targets metadata, specifically length and hashes about the target, matches across repositories. Custom targets metadata is exempted from this requirement.
4. If the targets metadata matches across repositories, return this metadata.
5. Otherwise, if the targets metadata do not match across repositories, or if none of the repositories signed metadata about the desired target, then take one of the following two actions.
5.1. If the "terminating" attribute is set to true, either report that the repositories do not agree on the target, or that none of them have signed for it.
5.2. Otherwise, process the next mapping following the same steps as above.

jhdalek55: Will probably take another run at this. It needs a few tweaks.

trishankkarthik:
@JustinCappos I've merged edits from @jhdalek55 and @ecordell.

ecordell:
I realized that we've missed documenting one of the big use-cases we originally discussed: allowing different teams to sign the same targets (with their own roles/thresholds) and letting clients pick and choose which teams' signatures to care about during metadata-verification (via the map file).

It's sort of encapsulated by the "AND relation" comment, but I think it's a useful use-case to call out explicitly. It can also be useful to contrast that with TAP3, where the publisher enforces a set of signatures from multiple roles (publisher-enforced), vs TAP4 where the client chooses which roles to care about (client-enforced).

JustinCappos:
Ah, okay!  So I was thinking we had encapsulated this in the first example,
but I now realize that this isn't true.  The first example has separate
repositories, whereas you want files on the same repo.  I think we could
either add a new example or adapt the current one to include this.

Do you have a preference?

ecordell:
I think I prefer a separate one, explicitly calling out the difference between TAP3 and TAP4...for implementers looking to use those features, the distinction between where/when signing policy is enforces is important.

trishankkarthik:
> I can make a PR with some copy for that if that helps?

Please do :)

vladimir-v-diaz:

JustinCappos:
On Fri, May 19, 2017 at 11:31 AM, Vladimir Diaz <notifications@github.com>
wrote:

> incorporated into TUF.
>

PR #21, opened 2017-05-11
Proposed TAP #9 - Mandatory metadata signing schemes

heartsucker:
Addresses issues covered in the conversation here: https://github.com/theupdateframework/tuf/issues/425

### DIFF
--- a/tap9.md
+++ b/tap9.md
@@ -1,0 +1,330 @@
+* TAP: 9
+* Title: Mandatory Metadata Signing Schemes
+* Version: 1
+* Last-Modified: 2017-05-11
+* Author: heartsucker
+* Status: Draft
+* Content-Type: text/markdown
+* Created: 20-Jan-2017
+
+# Abstract
+
+This was started by the discussion about standardizing signatures in [this
+GitHub issue](https://github.com/theupdateframework/tuf/issues/425#issuecomment-300792546)
+in the main TUF repository. Most of the information and arguments that arose
+from that discussion have been captured in this document.
+
+At present, metadata used in TUF is comprised of two sections: `signatures` and
+`signed`. The actual metadata is in the `signed` section while the `signatures`
+section contains cryptographic signatures of the data in `signed`. These
+cryptographic signatures include a `keyid`, `method`, and `signature`. An
+example piece of metadata is provided below.
+
+```python
+{
+  "signatures": [
+     {
+       "keyid": "588f02cf6907c3f199b201e845f5a458285e9ae956680e0e2fdcc7987306ef7c",
+       "method": "rsassa-pss",
+       "sig": "fce1a97b55076065c7dc7c78fb0206d5dd87db817ccc7bcf83c3cdf..."
+     }
+  ],
+  "signed": {
+    ...
+  }
+}
+```
+
+Multiple security researchers have pointed out that the `signatures` section
+exists in attacker controlled space and therefore should have the bare minimum
+amount of information in it necessary to validate a signature
+([reference 1](https://github.com/docker/notary/blob/b62321473789c704c02ee4ab68348821a5540d36/docs/resources/ncc_docker_notary_audit_2015_07_31.pdf)).
+This TAP explains why and how the `method` field will be moved into a role's
+definition such that any given key can use exactly one signature scheme to sign
+metadata.
+
+## Cure53 Audit of Advanced Telematic Systems GmbH
+
+The following is taken from [Cure53](https://cure53.de)'s review of the ATS
+implementation of TUF/Uptane.
+
+### ATS-02-003 Client: Signature method does not have to match key type (Info)
+
+It was found that it is not necessary for the method field of the signature to
+match the key type. The keyid field in the signatures' list is used to select
+the key which is then employed for verifying the signature. The keytype of the
+matching key is then used to determine the algorithm responsible for the
+signature verification. This process ignores the  method specified for the
+signature. While this does not pose an issue at the moment, if more signature
+schemes are being implemented in the future, the ability to control which
+scheme is used could potentially lead to security problems and risks.
+
+root.json signature example:
+
+```python
+{
+  "signatures": [{
+    "keyid": "8c7e2de8e96047a9d473e29c0d35daf5ec79b51addf62...<snip>",
+    "method": "rsassa-pss",
+    "sig": "oAsb1STplQf...<snip>...j2XsDvevw=="
+  }],
+  "signed": {
+    "keys": {
+      "20b1b4c6fd067c1b87c67b80890...<snip>...": {
+        "keytype": "RSA",
+        "keyval": { "public": "-----BEGIN PUBLIC...<snip>...\n" }
+      },
+      ...
+    },
+    ...
+}
+```
+
+It is recommended to reconsider whether the method field of a signature is
+actually needed. This especially holds when there is a mismatch between how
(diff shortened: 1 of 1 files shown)


tap9.md
```diff
@@ -0,0 +1,252 @@
...
+cryptographic signatures include a `keyid`, `method`, and `signature`. An
+example piece of metadata is provided below.
+
+```python
+{
+  "signatures": [
+     {
+       "keyid": "588f02cf6907c3f199b201e845f5a458285e9ae956680e0e2fdcc7987306ef7c",
+       "method": "rsassa-pss",
+       "signature": "fce1a97b55076065c7dc7c78fb0206d5dd87db817ccc7bcf83c3cdf..."
```
vladimir-v-diaz: The "signature" entry should be "sig"
See [section 4.2](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L443-L453) of the specification and [example](https://github.com/theupdateframework/tuf/blob/5e215ee223e6117c97b3407ee309daed66ddaac8/tests/repository_data/repository/metadata/timestamp.json#L2-L8) metadata.

tap9.md
```diff
@@ -0,0 +1,252 @@
...
+     }
+  ],
+  "signed": {
+    ...
+  }
+}
+```
+
+Multiple security researchers have pointed out that the `signatures` section
```
vladimir-v-diaz: Can you provide a reference to one of these audits, if possible?  It will be helpful to the
reader.

I know of one audit that is available online for review: https://github.com/docker/notary/blob/master/docs/resources/ncc_docker_notary_audit_2015_07_31.pdf

tap9.md
```diff
@@ -0,0 +1,252 @@
...
+such that any given key can use exactly one signature scheme to sign metadata.
+
+This was started by the discussion about standardizing signatures in [this
+GitHub issue](https://github.com/theupdateframework/tuf/issues/425#issuecomment-300792546)
+in the main TUF repository. Most of the information and arguments that arose
+from that discussion have been captured in this document.
+
+# Motivation
+
+Moving the `methods` field into the signed portion of the metadata removes one
```
vladimir-v-diaz: typo: "methods" should be "method."

tap9.md
```diff
@@ -0,0 +1,252 @@
...
+    ],
+    "keyval": {
+        "public": "-----BEGIN PUBLIC KEY-----<etc>"
+    }
+  },
+  ...
+}
+```
+
+When a signature is made, the `signature` object would still need to have a
```
vladimir-v-diaz: In the following example, assume that a developer has access to the same RSA key.

What if the "keys" entry was formatted as follows?

```python
"keys": {
  "0bde74244f238cd6d4e3839d35a5b248d6cde5759b760ebdbee45a386109a41a": {
    "keytype": "rsa",
    "scheme": "rsa-pkcs1-1.5",
    "keyval": {
        "public": "-----BEGIN PUBLIC KEY-----<etc>"
    },
{
  "349838deff9a9a9...": {
    "keytype": "rsa",
    "scheme": "rsassa-pss-sha256",
    "keyval": {
        "public": "-----BEGIN PUBLIC KEY-----<etc>"
    },
  },
  ...
```


And the "signatures" entry could list multiple signing keys like so:
```
{
  "signatures": [
     {
       "keyid": "349838deff9a9a9...",
       "sig": "fce1a97b55076065c7dc7c78fb0206d5dd87db817ccc7bcf83c3cdf..."
     },
    {
       "keyid": "0bde74244f238cd6d4e3839d35a5b248d6cde5759b760ebdbee45a386109a41a",
       "sig": "fce1a97b55076065c7dc7c78fb0206d5dd87db817ccc7bcf83c3cdf..."
  ],
  "signed": {
    ...
  }
}
```

tap9.md
heartsucker: I'll ask tomorrow in the office. I'm not sure what of our report we can share.

tap9.md
vladimir-v-diaz: Note that I edited my previous comment to change the "sig" listed by the "0bde7..." key, which should differ from the one provided by the "3498..." key.

heartsucker: I have addressed this in the latest commit.

tap9.md
```diff
@@ -0,0 +1,286 @@
...
+        "public": "-----BEGIN PUBLIC KEY-----<etc>"
+    }
+  },
+  ...
+}
+```
+
+This itself would be insufficient because when checking a threshold, a client
+would have have to check that each key ID calculated in the above way points to
+exactly one real key. We would have a calculate a second key ID that is
```
vladimir-v-diaz: Not that I'm in favor of this view, but I suppose a key in this case would be the KEYID entries in "keys."  In other words, there are two unique keys in the example above: one rsa key + 2 different signing schemes.

However, I agree that this can potentially be a source of confusion to users, who are likely to view a " real key" by its keytype, and not keytype + signing scheme.

tap9.md
heartsucker: We can't release the doc at the moment, but the relevant excerpt was placed into the TAP.

tap9.md
heartsucker: It would still need the check against the "real key ID" because the whole point of thresholds is that N entities sign the key, and allowing a single key to do multiple signatures means that an entity could provide all the signatures needed. For example:

Key with key ID `sha256(cjson(encoded)) = 123` has been placed in the metadata for the targets role with signing schemes `A`, `B`, and `C`. The new key IDs done via `sha256(cjson(scheme || encoded))` would be `456`, `abc`, and `def`. If the targets role has `threshold = 3`, then the single key `123` could sign it 3 times using "keys" `456`, `abc`, and `def`. To avoid this, we'd have to check that the `sha256(cjson(endcoded))` of each signature is unique which seems like an overly complicated step.

vladimir-v-diaz: That's correct.  This point was raised in issue #425 and I was under the impression that @JustinCappos views "keys" `456`, `abc`, and `def` as unique, even though they all use the same `123` key.

vladimir-v-diaz: Link to discussion:
https://github.com/theupdateframework/tuf/issues/425#issuecomment-300300159

vladimir-v-diaz:
TAPs that have been assigned a TAP number will now be merged into the "master" branch.  This does not mean that the merged TAP has been accepted, only that the proposed change (with potential tweaks) is likely to be incorporated into TUF.

Thanks for submitting TAP 9!

JustinCappos:
On Fri, May 19, 2017 at 11:22 AM, Vladimir Diaz <notifications@github.com>
wrote:

> incorporated into TUF.
>

JustinCappos:
@endophage: I'd love to have your thoughts on this TAP.

tap9.md
endophage: I would stick with the original solution. The RSA keys themselves are likely to be bulk of the data inside a root.json so not duplicating them seems beneficial.

endophage:
@JustinCappos I'd like to think that every client matches the signing scheme against the key type regardless of where those pieces of data are located so the premise for this issue seems to be "in a poorly implemented client ..." which isn't the strongest of reasons.

That said, I have no issues with the change and agree that it increases this aspect of security in those poorly implemented clients.

JustinCappos:
Okay, does anyone else have a thought / concern?  Any objections with moving this to accepted?

PR #30, opened 2017-05-24
TAP 7: Simplify and Augment: Wrapper Module, Fewer Return Values, Description of Data Provided

awwad:
The primary changes in this PR to the way that TAP 7 works are:
- Shift from a model involving command line instructions to a model involving a "Wrapper" module that is expected to implement three functions. Also provide a working example of a Wrapper for the TUF Reference Implementation.
- Move from many return codes to just 3 options: Success 0, Failure 1, or (for debugging purposes) Unexpected Error 2
- Describe the data that will be provided by the Tester to the (Wrapper and) Updater
- Solidify terminology
- Make clear that other implementations aren't expected to be in Python
- TAP 4 support (optional)
- Trim some config file options (root keys and thresholds) and add some others (TAP 4, delegations, mirrors)

Commit summaries can provide more details.


awwad:
Okay, I think this is worth glancing over now, @JustinCappos, to see if the general scheme is satisfactory. (Also, @vladimir-v-diaz, LMK if it looks fishy compared to what we've discussed?)

Note that I haven't very substantially modified the Abstract, Motivation, or Rationale sections.

tap7.md
```diff
@@ -63,47 +64,263 @@ that claim becomes stronger.  Consequently, any
...
+will be expected to reject untrustworthy metadata and targets and accept
+trustworthy metadata and targets.
+
+The **Wrapper** mediates communication between the Tester and Updater, adapting
+metadata and communicating it in the way the Updater expects. An
+individual Updater will need a custom Wrapper written for the Tester to use to
+communicate with it. This will need to involve at least a few lines of Python.
+In order for the Tester to interact with the Updater implementation, a Wrapper
+around that implementation will need to support as an interface to the Tester
+the 3-5 functions listed below.
```
vladimir-v-diaz: So far it sounds as if developers are required to create the Wrapper from scratch, but I think we'd like to provide a skeleton of the wrapper and have developers sort of fill in the blanks for the initialize_updater(), update_client(), etc. functionality.

tap7.md
```diff
@@ -63,47 +64,263 @@ that claim becomes stronger.  Consequently, any
...
+
+
+## Wrapper Specification
+
+The Wrapper must implement at least the first three functions specified
+[below](#wrapper_functions).
```
vladimir-v-diaz: This link doesn't work on my end.

tap7.md
```diff
@@ -163,276 +463,162 @@ updater = tuf.client.updater.Updater('repository', repository_mirrors)
...
+#### Key Restrictions
+
+Suppose, for example, that the Updater implementation has the following restrictions:
 ```
-(1) metadata is encoded in DER (rather than JSON)
-(2) only Ed25519 keys are used and listed in metadata
-(3) the Root file must be signed by 1 out of 2 keys (i.e., threshold of 1)
+- only Ed25519 keys are used and listed in metadata
+- only exactly 2 keys are supported for Root metadata, and only a threshold
+    of 1  (((Is this necessary?)))
```
vladimir-v-diaz: The conformance tool can potentially test for a Root threshold that an implementation doesn't support.  For instance, the tool attempts to test for multiple Root keys, but an implementation fixes the number of Root keys to 1.

tap7.md
```diff
@@ -114,32 +331,115 @@ defend against the following attacks and weaknesses:
...
+
+
+#### Case Multipliers
+When a Case Set below lists one of these multipliers, the number of tests of
+that case set will need to be multiplied in the listed way.
+- **Per Role**
+    - Run the case at least five different ways, where the offending metadata
+      is a different role: Timestamp, Snapshot, Target, Delegated Target
+      (((depth x > 1, x < 5?))), Root.
+- **Mirrors**
```
vladimir-v-diaz: What if an implementation doesn't care about mirrors and doesn't bother to support it?

tap7.md
```diff
@@ -114,32 +331,115 @@ defend against the following attacks and weaknesses:
...
+metadata has been validated and the target obtained matches that now-validated
+metadata.
+
+
+#### Case Multipliers
+When a Case Set below lists one of these multipliers, the number of tests of
+that case set will need to be multiplied in the listed way.
+- **Per Role**
+    - Run the case at least five different ways, where the offending metadata
+      is a different role: Timestamp, Snapshot, Target, Delegated Target
```
vladimir-v-diaz: What if an implementation doesn't use nor support delegated roles.  I think Flynn's Go implementation falls into this case.

tap7.md
awwad: True: we should provide a skeleton. It wouldn't be much larger than the specification text. At the same time, I wouldn't want to muddy the specification itself with a bunch of imports and extra detritus, so what do you think: where should we put the skeleton? .py file in a folder in the taps repository, linked to here?

tap7.md
awwad: I used _ instead of - in the links in the document. Fixing.

tap7.md
awwad: Not supporting delegated roles is OK per the spec, right?

I'll fix this by noting that delegated roles are optional in the Per Role multiplier, and adding support for delegated roles to the sample config file.

tap7.md
awwad: True. (Not supporting mirrors is OK per the spec, right?)

I'll fix this by noting in the Mirrors multiplier that if multiple mirrors are not supported, the tests will instead be with 1 good mirror or 1 bad mirror. I'll also add support for delegated roles to the sample config file.

tap7.md
vladimir-v-diaz: Yup, I think the TAPs repository would be good place to store resources used by the TAPs.  These resources can be images, python modules, example metadata, etc.

tap7.md
awwad: Unless I misunderstand, I don't think that's helpful. (I'll poke you tomorrow and ask to make sure I do understand.) We shouldn't really worry about testing conformance with things the TUF spec doesn't require in the first place.

awwad: To help settle the question of whether or not threshold and key count specification is necessary for Tester configuration: are you aware of any existing need for this, Vlad?

(@JustinCappos: wrt your question yesterday)

awwad: We talked about this briefly. We're not currently aware of any implementation that doesn't support thresholds > 1 or doesn't support multiple root keys<sup>1</sup>.

We've decided to include a sentence along the lines of "Other configuration options may be added to handle other constraints, if implementers reach out with their needs - for example, if a particular implementation doesn't support key thresholds > 1, an option could be added to this configuration file."

<sup>1<sub>TBC, I mean multiple keys *listed* in root / treated as root keys simultaneously. Allowing multiple *signatures* is a required feature without which key rotation breaks).</sub></sup>

vladimir-v-diaz: I don't know for certain, but Docker Content Trust might only allow one Root key per repository.
See https://docs.docker.com/engine/security/trust/trust_key_mng/

Notary's [test case](https://github.com/docker/notary/blob/bdf934353d4dc5117d1d800d71ded2a34ebdfd6e/tuf/data/root_test.go#L169-L203) for the Root role appears to only consider a threshold of 1.

I can imagine other simpler adoptions that might only bother to support one Root key.

JustinCappos:
>
> We've decided to include a sentence along the lines of "Other
> configuration options may be added to handle other constraints, if
> implementers reach out with their needs - for example, if a particular
> implementation doesn't support key thresholds > 1, an option could be added
> to this configuration file."


Hold a moment.  A TAP is supposed to say what an implementer needs to do to
implement TUF correctly.  It's strange to say something like 'we may add
other options', unless you need them to future proof their implementation
in some way.

vladimir-v-diaz:
I think that the conformance tests will likely need updating as the specification evolves.  They might also need updating in the future to include additional test cases.

How should we state that along with the conformance tests, additional restriction options may be supported in the future by the tool's configuration file?

Note that these options are for configuration settings that are not shared equally across all implementations, although they are still allowed by the specification.  For example, the Go implementation might only support ECDSA keys, whereas another might support Ed25519 and RSA keys.

We are not yet certain if a restriction option for the number of Root keys is needed.  Should we go ahead and mention this option in TAP 7, or state that it can be supported later if the need arises?

JustinCappos:
Supporting different key types seems not too hard to handle.  After all,
the implementer may need to have custom signing / metadata parsing
routines.  But I don't think we should define things in the TAP that we
aren't sure we will add.  If we later have a need for another configuration
option, etc. we could add it.  But let's not stick things in the TAP we may
never use.

awwad:
Not including something we're not sure we'll need in the TAP sounds right to me. I guess I only mean to communicate flexibility on these limitations in some way to those implementers reading the TAP. I suppose it'll come up in discussion. We'll exclude the statement from the TAP itself.

vladimir-v-diaz:
We should discuss whether providing only the different sets of metadata, along with their expected outcome, is sufficient for a developer to test an implementation.  All of the update conditions and their outcomes should be possible to test with only metadata (with the exception of the slow retrieval attack). 

I wonder if this TAP is unnecessarily complex for our goal, which is to provide a way for developers to verify that an implementation follows the specification.  All a developer needs to do is load our metadata (potentially needing to convert it into another format) into their system and note the outcome.  From there, the developer can verify whether the outcome matches our expected result.

I'm not sure that requiring developers to generate a wrapper script or modify their implementation to work with it is making it easy for them.  Is it necessary to dedicate an entire TAP to explaining how to use a tool that loads different sets of metadata and spits out an answer to the question, "Does this implementation follow the specification?"

However, there might be a good need for a tool that generates test metadata according to the restrictions of the implementation.

JustinCappos:
>
> We should discuss whether providing only the different sets of metadata,
> along with their expected outcome, is sufficient for a developer to test an
> implementation. All of the update conditions and their outcomes should be
> possible to test with only metadata (with the exception of the slow
> retrieval attack).


Well, you also need to know how to set up the repo(s), what the initial
state of the client should be, what target the client should try to
download, and what the result should be.  This is basically exactly what I
see TAP 7 as specifying.

The only added thing is how this is communicated to the program that does
the tests.  I suggest having a few simple function calls as a sane way to
do so.

awwad:
I've added the code for a fully functional Wrapper for TUF's Reference Implementation's updater, which should help people understand what is required.

I've also added an _extremely minimal sample_ of what a Tester would do.

Ultimately, I'd split these into separate pull requests, but for the sake of discussion at this stage, I've kept things all here.

awwad:
I think this is ready to be reviewed again, @JustinCappos and @vladimir-v-diaz.

awwad:
Side question: I think I should probably remove the actual example file, since the text of it is now in the TAP. I'll do that if it seems sensible to you.

vladimir-v-diaz:
The title should probably change, there is more involved in these changes than removal of the subprocess requirement.

tap7.md
```diff
@@ -63,47 +64,320 @@ that claim becomes stronger.  Consequently, any
...
+
+
+A skeleton that can be filled in by implementers is provided
+[here](tap7_resources/tap7_wrapper_skeleton.py).
+Also see [Example Wrapper](#example-wrapper) below for a functioning example of
+the Wrapper module - in this case, a Wrapper enabling the Conformance Tester to
+test the pre-TAP4 TUF Reference Implementation.
+
+
+## Test Specification
```
vladimir-v-diaz: I was expecting  a `Tester Specification` heading, to match the `Wrapper Specification` section.

tap7.md
```diff
@@ -1,29 +1,29 @@
 * TAP: 7
 * Title: Conformance testing
-* Version: 1
-* Last-Modified: 17-May-2017
+* Version: 3
```
vladimir-v-diaz: Why was this changed to 3?

Perhaps this is more a question about the version number changes to this field:  Is there a way for us to track or document the history of these version number changes?

tap7.md
```diff
@@ -63,47 +64,320 @@ that claim becomes stronger.  Consequently, any
...
+including error codes that signal that particular attacks have been detected --
+whether or not the Updater is conformant with the TUF specification. In
+general, these constitute a battery of attacks against which the Updater should
+be resilient.
+
+It should be noted, however, that the implementation does not necessarily have
+to be the Updater used in production. It only needs to function as defined in
+this TAP for conformance testing, though it is expected that the behavior be
+the same at a high level -- for example, the validity of metadata should be
+determined the same way. ((TODO: This paragraph still seems wordy. Unnecessary?))
```
vladimir-v-diaz: Do you think the content of the paragraph is unnecessary?

@JustinCappos requested that I include in the TAP.  For one, it makes it clear that we are not requiring developers to change their updater (at least the one used in production), and a second reason is to prevent a production implementation from accidentally shipping with conformance testing features enabled, which have led to security issues in the past. 

Justin can explain more.

tap7.md
awwad: I suppose the blame view is the best way to see this. https://github.com/awwad/taps/blame/1ef1c8066b78c2f31dcc538ede852ad6878d4098/tap7.md for example. Then you can click the little array-of-boxes (stack-of-papers?) button to see the previous blame.

tap7.md
awwad: I added the "though it is expected that the behavior be the same at a high level" bit to try to reduce the impression that you can have entirely different behavior in testing mode, but it feels wordy. I was hoping for something succinct that conveys that things may have to work *slightly* differently in test mode, and that that's okay.

awwad: Maybe:

```If the behavior necessary to provide the Conformance Tester with what it needs to judge conformance is slightly different from the usual behavior of the Updater, that is OK. For example, if errors are usually ignored rather than producing any return value, that's something that the Wrapper module can adjust.```

File: tap7.md (hunk 6 above)
Comment (awwad) [inline tap7.md]: External parties have to implement Wrappers with specific functions provided that behave in specific ways, whereas the Tester is something we'd provide, so it didn't seem to need the same level of attention. I thought that the key information would be what the tests look like. 

There's also a sample of some Tester code linked to elsewhere, and a description of the high level behavior of the Tester in section Executing Conformance Testing.

Do you think we should specify more, and put it in its own section here?

File: tap7.md (hunk 7 above)
Comment (vladimir-v-diaz) [inline tap7.md]: The reader can go digging through the blame view, but that seems like a lot of work.  Ideally, the version number changes should be documented in the TAP itself.

[TAP 1: TAP purpose and guidelines](https://github.com/theupdateframework/taps/blob/master/tap1.md) states that the `Post-History` field should list the dates in which new versions of a TAP are posted to the mailing list.  I'm assuming these versions of the TAP correspond to the `Version` field.

I guess the reader can go back and look at the documented changes on the mailing list, but that wouldn't work in our case since the changes are happening on GitHub.

File: tap7.md (hunk 6 above)
Comment (vladimir-v-diaz) [inline tap7.md]: I was only expecting that the title of the heading be `Tester Specification`, to follow the names used in the TAP up to that point.

If you don't think `Tester Specification` captures the spirit of that section, then consider making it clear in the title that it it goes over the tests used by the Tester.

File: tap7.md (hunk 7 above)
Comment (awwad) [inline tap7.md]: I guess I was thinking of a much more granular version history. It looks like every TAP is currently listed at version 1 (or in the case of TAP 2 has no version). I'll make this version 1 again to fit that model.

Question while we're on the topic: from the description of Post-History, it's not clear to me which of these two Post-History should look like: Version X lists the posted date for version X in Post-History, or Version X lists the posted dates for all versions V<=X in Post-History.

```
Post-History: v1: 14-Aug-2001, v2: 01-Jan-2002, v3: 02-Feb-2003
```
?

Comment (vladimir-v-diaz) [inline tap7.md]: Good question.  I think a single date is expected according to the description, but the posted dates for all versions would be more useful.

File: tap7.md [hunk 9]
```diff
@@ -63,47 +64,320 @@ that claim becomes stronger.  Consequently, any
...
+          away.
+
+    - Returns:
+        An integer describing the result of the attempted update. This value is
+        what the Tester is ultimately testing.
+
+        ```
+        return value     outcome
+        -----------      ------
+        0                SUCCESS: target identified by target_filepath has been
```
vladimir-v-diaz: A blank line after each outcome might help readability.

tap7.md
awwad: I've reorganized most of the Specification portion of the document into these sections (commit 5a7c728), so that it'll hopefully be more intuitive:

- Configuration File Specification
- Tester Specification (incl. Test Battery)
- Wrapper Specification (incl. Example Wrapper)

I hope that makes more sense?

tap7.md
awwad: I went with this in commit 79ff09c:

```
The behavior necessary to provide the Conformance Tester with what it needs to
judge conformance may be slightly different from the usual or production
behavior of the Updater, resulting in a need for a testing mode, or logic in
the Wrapper to interpret behavior. For example, if errors are usually ignored
rather than producing any return value, that's something that may be adjusted
by using test-mode-specific code in the Updater, or post-hoc by the Wrapper
module. The validation behavior during testing should not vary significantly
from that in production so that test results can represent real Updater
performance.
```

vladimir-v-diaz: That works for me!

tap7.md
awwad: K, added in e42f27c

awwad:
I added a very brief blurb in the TAP on why we're going with two expected return values (from the Wrapper's `updateinstead of many. Here's more of my thinking, for reference.

**Argument 1: We don't need specific error codes.**
I can't think of a case where we actually need specific error codes to test behavior. We're testing updaters, so we can always ask the question "Did it update without manipulation X and not update with manipulation X?"

**Argument 2: Many error codes may be onerous.**
To demand that an updater issue distinct errors that percolate up through the code may be an onerous requirement on an implementer, and we want people to not hesitate to use the conformance tester. I'd rather we do extra work in one place to construct good test cases than every implementer work to adapt their updaters to issue error codes that exactly match our expectations.

**Argument 2.5: Many error codes encourages test mode divergence from live code.**
While test mode could be required for some implementations, the more logic we demand of it, the further it could drift from their live, production code. In the ideal solution, a Wrapper module is thin and it is possible to have conformance tests run with every build of live code, but that can only work if we can test live code or test with only minimal modifications, preferably contained in a static Wrapper module.

**Argument 3: Many error codes requires detailed and precise specification of behavior.**
The expected behavior if an updater contacts one mirror with invalid metadata and one mirror with valid metadata is simply to update, right? Even the reference implementation throws errors from other mirrors away if one mirror works out. Is that what we expect, or do we expect to get an array of responses per mirror?
A dictionary of responses per mirror means we're specifying that format. It's one more hoop for implementers to jump through.
In the absence of a successful update from one mirror, it might be perfectly sensible for a given Updater not to protest if it finds invalid metadata, but just not accept it and try again a few times or contact the most authoritative source.
I'm sure there will be debates about which error code is more appropriate in a particular scenario.

vladimir-v-diaz:
Note: @JustinCappos requested that we merge this pull request.

PR #41, opened 2017-06-05
Update TAP 7's Last-Modified and Post-History

vladimir-v-diaz:
Match TAP 1's approach, where the Post-History is not empty if it equals the creation date.
https://github.com/theupdateframework/taps/blob/master/tap1.md


vladimir-v-diaz:
Correction: In Post-History, I list the date when TAP 7 was posted to the TUF mailing list.

PR #42, opened 2017-06-06
TAP 7: Improve Rationale section - cut and clarify

awwad:
This PR rewrites the Rationale section. Among the changes:

- removal of reference to error codes
- trimming some pronouns
- connecting conjunctions
- being meticulously explicit
- moving the 2-vs-many return codes rationale to the Rationale section (rather than the Wrapper Specification)

### DIFF
--- a/tap7.md
+++ b/tap7.md
@@ -77,8 +77,8 @@ will pass in sets of metadata and target files. In response, the wrapped
 Updater will provide return values indicating successful update or failure to
 update, and the Tester will compare these to the expected values for each test
 data set. In general, these data sets constitute a battery of attacks against
-which the Updater should be resilient, contrasted by controls which should
-succeed.
+which the Updater should be resilient, contrasted by positive control cases
+which should succeed.
 
 Note that the Tester will only expect a matching success-or-failure code
 instead of a specific code indicating the type of attack detected or error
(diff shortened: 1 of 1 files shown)


awwad:
While I think this is adequately explicit now, I worry that the TAP may be a bit redundant. Thoughts?

tap7.md
```diff
@@ -53,30 +53,44 @@ an important step in checking if an implementation is TUF-conformant.
...
+
+Therefore, This TAP prescribes that an implementation of a client updater
+("Updater") have a corresponding custom wrapper module ("Wrapper") that
+implements a common set of functions defined in this document. These Wrapper
+functions will be called by a general TUF Conformance Tester ("Tester"), which
+will pass in sets of metadata and target files. In response, the wrapped
+Updater will provide return values indicating successful update or failure to
+update, and the Tester will compare these to the expected values for each test
+data set. In general, these data sets constitute a battery of attacks against
+which the Updater should be resilient, contrasted by controls which should
```
vladimir-v-diaz: Should "controls" be "positive control cases?"

vladimir-v-diaz: Should "controls" be "positive control cases?"

awwad: Probably. I worried a bit about saying "controls" because I wasn't sure it'd be clear to people. If you think "positive control cases" is clearer, then that's great. Fixed in 58b8893.

PR #44, opened 2017-06-06
TAP 7: Clarify abstract

awwad:


tap7.md
```diff
@@ -21,10 +21,14 @@ and blocking known updater attacks) these unit tests only work within the
 parameters of the reference implementation. This is problematic due to the
 diversity of TUF implementations.
 
-To achieve the goal of testing diverse TUF implementations, this specification
-describes the design of a common testing tool for TUF conformance.  The goals
-are to interoperate with implementations in diverse languages and environments,
-while having a low burden on TUF implementers to the tool in their environment.
+This specification describes the design of a common testing tool for TUF
```
vladimir-v-diaz: Should it be "This proposal" ?  It's confusing to use the word "specification" here, which can be mistaken with the TUF specification.

vladimir-v-diaz: Should it be "This proposal" ? It's confusing to use the word "specification" here, which can be mistaken with the TUF specification.

awwad: Fixed. I think I inherited that there.

PR #46, opened 2017-06-07
TAP 7: Full Re-read Revisions

awwad:
Various edits made during a final read-through.

Changes in Round 1:
- Remove mention of number of root keys and root threshold in config
- Note that config options will be specified in Tester documentation
- initialize_updater may be called per test / test series
- handle multiple initialize_updater calls in wrapper example
- adjust appearance of some docstrings
- correct typos
- shift from 'you' to 'the Wrapper' where possible
- clarify how the keys argument should be interpreted in Wrapper functions
- add some links

Changes in Round 2 (based on Justin's feedback):
- rename Wrapper functions to make them more readily grokked
- move all repo setup from first Wrapper function to second
- rewrite Wrapper intro section so it's clear it's just an intro, and more understandable

### DIFF
--- a/tap7.md
+++ b/tap7.md
@@ -239,7 +239,7 @@ tap4-support: false
 The Wrapper must implement three functions. These are specified in detail
 [below](#wrapper-functions). In brief, the Tester will call the Wrapper's
 `set_up_initial_client_metadata` and `set_up_repositories` functions to assign
-initial trusted metadata to the client and put metadata and targets on the
+initial trusted metadata to the client, and put metadata and targets on the
 repository, respectively. After this, the Tester will call the Wrapper's
 `attempt_client_update` function to perform the test itself, instructing the
 client to try to update. The Tester will judge the correctness of the result
@@ -604,7 +604,7 @@ TUF Reference Implementation. (This can also be seen
     # Copy the provided metadata into a directory that we'll host.
     if os.path.exists('hosted'):
       shutil.rmtree('hosted')
-    assert os.path.exists(test_data_dir + '/test_repo'), 'Invalid ' + \
+    assert os.path.exists(test_data_dir + '/test_repo'), 'Invalid ' \
         'test_data_dir - we expect a test_repo directory.'
     shutil.copytree(test_data_dir + '/test_repo', 'hosted')
 
--- a/tap7.md
+++ b/tap7.md
@@ -538,6 +538,9 @@ TUF Reference Implementation. (This can also be seen
     global updater
     global server_process
 
+    # End hosting from any previous test.
+    kill_server()
+
     # Initialize the Updater implementation. We'll put trusted client files in
     # directory 'client', copying some of them from the provided metadata.
     tuf.settings.repositories_directory = 'client' # where client stores repo info
@@ -676,9 +679,11 @@ TUF Reference Implementation. (This can also be seen
     Kills the forked process that is hosting the repositories via Python's
     simple HTTP server
     """
+    global server_process
     if server_process is not None:
       print('Killing server process with pid: ' + str(server_process.pid))
       server_process.kill()
+      server_process = None
 ```
 
 
--- a/tap7.md
+++ b/tap7.md
@@ -460,16 +460,15 @@ is available, and an [example is available below](#example-wrapper) as well.
 
 - 3: **`attempt_client_update(target_filepath)`**:
     - Purpose:
-        Refreshes metadata and causes the client to attempt to (obtain and)
-        validate a particular target,
-        along with all metadata required to do so in a secure manner conforming to
-        the TUF specification.
+        Cause the client to attempt to refresh metadata from the repository
+        and obtain and validate a particular target, in a secure manner
+        conforming to the TUF specification.
 
         This function will have to translate Updater behavior/output into the
-        return values (below) that the Tester expects, based on
-        whether or not the Updater detects a particular attack. `update_client`
-        must return the appropriate code to the Tester, which will evaluate them
-        against what it expects.
+        return values (below) that the Tester expects, based on whether or not
+        the Updater updates successfully. It must return the appropriate value
+        to the Tester, which will evaluate this return value against what it
+        expects.
 
     - Arguments:
         - `target_filepath`:
@@ -521,7 +520,7 @@ TUF Reference Implementation. (This can also be seen
     This is an example of a Wrapper module, which enables the Conformance Tester
     (as described in TUF TAP 7) to communicate with a particular TUF-conformant
     Updater implementation - in this case, the pre-TAP4 TUF Reference
-    Implementation (configuration file option tap4-support: false).
+    Implementation (configuration file option tap_4_support: false).
 
     The Conformance Tester will call the functions listed here in order to
     perform the tests necessary to ascertain the conformance of the Updater to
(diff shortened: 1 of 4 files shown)


tap7.md
```diff
@@ -230,42 +230,42 @@ mirror-support: false
...
- format the Updater expects, potentially re-signing metadata if the Updater
- will expect signatures over a different format
- - If the Updater's communication model involves different synchronization
- (e.g. server push vs client pull), the update_client() Wrapper function will
- need to bridge this; for example, it may need to wait and collect results from
- some separate process.
+The Wrapper must implement three functions. These are specified in detail
+[below](#wrapper-functions). In brief, the Tester will call the Wrapper's
+`set_up_initial_client_metadata` and `set_up_repositories` functions to assign
+initial trusted metadata to the client and put metadata and targets on the
```
vladimir-v-diaz: This is fine, but it might be even clearer to include a comma like so:

> ... functions to assign initial trusted metadata to the client, and put metadata and targets on the repository,
> respectively.

I find that the comma makes it easier to match description with function.

tap7.md
```diff
@@ -561,12 +583,30 @@ TUF Reference Implementation. (This can also be seen
...
+    # End hosting from any previous test.
+    kill_server()
+
     # Repository Setup
 
     # Copy the provided metadata into a directory that we'll host.
     if os.path.exists('hosted'):
       shutil.rmtree('hosted')
-    shutil.copytree(trusted_data_dir + '/test_repo', 'hosted')
+    assert os.path.exists(test_data_dir + '/test_repo'), 'Invalid ' + \
```
vladimir-v-diaz: I think the `+ \` isn't needed.

tap7_resources/tap7_wrapper_example.py
```diff
@@ -74,12 +72,30 @@ def initialize_updater(trusted_data_dir, keys, instructions):
...
+  # End hosting from any previous test.
+  kill_server()
+
   # Repository Setup
 
   # Copy the provided metadata into a directory that we'll host.
   if os.path.exists('hosted'):
     shutil.rmtree('hosted')
-  shutil.copytree(trusted_data_dir + '/test_repo', 'hosted')
+  assert os.path.exists(test_data_dir + '/test_repo'), 'Invalid ' + \
```
vladimir-v-diaz: Please also verify that the `+ \` is not needed here.

tap7_resources/tap7_wrapper_skeleton.py
```diff
@@ -45,15 +45,21 @@ def initialize_updater(trusted_data_dir, keys, instructions):
           file.
 
           This structure allows for optional multi-repository support per
-          [TAP 4](tap4.md). If TAP 4 is not supported (See
-          [Configuration File](#configuration-file)), then map.json will be
+          (tap4.md). If TAP 4 is not supported (See
+          tap7.md#configuration-file-specification), then map.json will be
```
vladimir-v-diaz: Haven't looked at the formatted doc, but isn't there brackets missing here?

tap7_resources/tap7_wrapper_skeleton.py
```diff
@@ -45,15 +45,21 @@ def initialize_updater(trusted_data_dir, keys, instructions):
...
           excluded, and there will only be one repository directory, named
           test_repo.
 
           The data provided for
-          initialize_updater should be treated as already validated.
+          set_up_initial_client_metadata should be treated as already validated.
```
vladimir-v-diaz: I wonder if we should have a test where the local Root isn't actually valid (e.g., bad signatures).  Not sure if that would be needed, would need to verify again with the specification...

tap7_resources/tap7_wrapper_example.py
awwad: Because this is a plain assert (the strings aren't contained in anything), the \ is needed. The + isn't, but it's explicit, I suppose.

tap7_resources/tap7_wrapper_skeleton.py
awwad: The markdown doc has this saved differently. Since this isn't a .md file, I've just written a relative link, which people can type in. I didn't really want to have the whole long URL, but we could, if you think it's worth it.

tap7_resources/tap7_wrapper_skeleton.py
awwad: We'd have to think about that. Something like that is probably worth doing, in which case maybe the wording here could be "trustworthy" rather than validated.... It starts to put too fine a point on the wording, I guess.

awwad: I'm unsure.... Do we want to mandate a certain behavior with regard to the initial metadata that ships with a client? That might go above and beyond. Does the spec say anything about how to do this?

tap7_resources/tap7_wrapper_example.py
vladimir-v-diaz: Hmm, I thought if the strings were within parenthesis (os.path.exists in this case), it is using Python's implied line continuation.  I guess some testing is needed from my side.

tap7.md
awwad: Because this is a plain assert (the strings aren't contained in anything), the \ is needed. The + isn't, but it's explicit, I suppose. I'll get rid of the +, I guess.

tap7.md
awwad: Fixing

tap7_resources/tap7_wrapper_example.py
awwad: Note that the strings aren't in parentheses.

tap7_resources/tap7_wrapper_skeleton.py
vladimir-v-diaz: The specification does say that Root must be signed with a valid threshold of keys...

https://github.com/theupdateframework/tuf/blob/8ddc62c43ff05fee53776e54985071bf83a00c89/docs/tuf-spec.txt#L950-L957

tap7_resources/tap7_wrapper_example.py
vladimir-v-diaz: Ahhh, I see it now!  The 'Invalid ..." string comes after the comma.

tap7_resources/tap7_wrapper_skeleton.py
vladimir-v-diaz: After reviewing the detailed workflow with Sebastien, the detailed workflow could  do without this step, and possibly use a different word for "previous" (or the currently trusted version of Root).

awwad: I think that step 0.1 in the linked workflow is not necessary in the specification and could be misleading. The check we're performing is on an already-trusted root file.

Either the very same check was already performed on the same file by the previous run of the Updater, in part (2) of step 1.3, or this root file is data that initially shipped with the updater, in which case we trust it implicitly. There is no gain from this validation.

awwad: In summary, my opinion is that we leave this text here as is, and we remove step 0.1 of the update workflow.

vladimir-v-diaz: I agree that step 0.1 should be removed, given steps 0 and 1.3.

Issue #36, opened 2017-05-31
Discussion of TAP 7: Conformance testing


vladimir-v-diaz:
TAP 7 is available at https://github.com/theupdateframework/taps/blob/master/tap7.md

Pull requests related to TAP 7:
#19, #30, #40, #41, #42, #43, #44, #55 

Additional discussion of TAP 7 on the TUF mailing list:
https://groups.google.com/d/msg/theupdateframework/6XHRPTAQdbI/oFW3Ox8YBQAJ

awwad:
Additional pull requests related to TAP 7:
#30 
#40 
#41 
#42 
#43 
#44 

(The pull-requests-for-everything model among writers of the spec before the spec is accepted seems a bit excessive, btw.)

heartsucker:
Ciao, y'all. Responding to your email here to aid in public discussion.

I think the idea of a conformance tester is appropriate, however I think there is a problem of near limitless configuration issues. Some of this was touched upon with the idea of `.tuf-tester.yml` to configure the conformance tester. Some of the options there make complete sense, but with the possibility of moving away from strictly using JSON as the data interchange, I think it would be infeasible to have all possible options for example.

Imagine these options that the conformance tester would have to support.

- Data interchange:
  - JSON
  - ASN.1
  - Protobuf
- Signing Key Types (not to mention how they might be named `rsa-ssa-pss` or `rsa-ssa-pss-sha256` or `RSASSA_PSS_SHA256`)
  - RSA
     - PKCS#1
       - over SHA256
       - over SHA512
       - etc.
     - SSA-PSS
       - over SHA256
       - over SHA512
       - etc.
  - Ed25519
  - ECDSA
- Field names
  - `type` vs. `_type` (I really think the spec shouldn't use an underscore here)
  - `sig` vs. `signature` (an implementation may choose to be verbose for clarity)
- Field Values
   - `Root` vs. `root` (again, I think capitalization here is unnecessary)
- Hash maps vs. lists
  - The `signatures` value is a list with unique values, so it might as well be a hash map of `keyid` to `sig`
- Usage of mirrors role (TAP4 #33)
- Possible different definitions of "canonical" for the given data interchange format
  - CJSON (https://github.com/theupdateframework/tuf/issues/457)
  - ASN.1 -> DER, BER, etc.
- Multi-role delegations (TAP3 #32)
- Signature / hash encoding
  - base64
  - base32
  - hex
- Key encoding
  - RSA
    - PEM/PKCS1
[...]

PR #48, opened 2017-06-27
TAP 7: Convert from Tester spec to test data spec

awwad:
Rewrite TAP 7 as a test data specification (instead of Tester specification) based on discussion occurring here: #36

This is a response to concerns from several parties that it could be too hard to
integrate and use the previous TAP 7 design (version 1). The version here has been increased to version 2.


awwad:
I think this can be reviewed now.

heartsucker:
I think providing already signed metadata isn't descriptive enough.

Here's the format I envisioned: https://gist.github.com/heartsucker/a51f9830a1e9808da37dae32fcdc2ab5

Basically it fully describes the way the metadata should look on the way out. It's easier to say "create a bad hash" than to give a hash and a target and expect the impelementor to check that the hash is bad then encode that logic in how they create their test case. Or "list the size incorrectly" rather than have the targets metadata have actual size 1337 and the snapshot report it's size as 1336 and expect the test implementor to catch that.

My format also doesn't explicitly provide keys, just "key indexes" in some imaginary repo of pregenerated keys. If you give only ed25519 or 2048 bit RSA keys, and the implementer uses a curve other than curve25519 or force requires 8192 bit RSA, then they might not be able to interop easily and the impelementor would already be forced to diverge from TAP 7 just to get it to work.

So the format should be metadata describing the repo (metadata / targets) and not the actual metadata itself. Also means you could write a python tool that generates this test-vector-metadata and then have it dump it to JSON, XML, etc. very easily.

awwad:
Our thinking is that implementers would convert the provided metadata -- rather than deconstructing it, grokking any modifications that might have been made to it, and recreating it from scratch in your format.

The description provided with the test case should be thorough enough that hypothetically you could construct your own metadata, sure, but not in an automated way. That actually seems more constrained. Every time some new thing to test occurs to us, we'd need to add some property to the JSON and you'd need to change your parser to support it.

What's the issue with conversion?

heartsucker:
> Our thinking is that implementors would convert the provided metadata -- rather than deconstructing it, grokking any modifications that might have been made to it, and recreating it from scratch in their format.

If one of the test cases is "delegation A, a sub delegation of B, a delegation of targets was signed with the keys of delegation C, a delegation of targets", parsing those directives out of the metadata is basically impossible. There are infinite permutations, and I would rather have input as *what* to do, not an example of the output to reverse engineer programmatically. Or since you said the text description would explain what to do, if you're given a written example with enough detail to determine how to build the test case, there's really no point in providing the metadata in the first place. Further, structured directives in the form of JSON (or whatever) is going to be far less error prone than some human language.

For example, this is completely unambiguous:

```json
"sign": [
    {"key_index": 0, "bad_sig": true},
    {"key_index": 0, "bad_sig": false},
    {"key_index": 1, "bad_sig": false},
]
```

But saying it in English could lead to errors. Now multiply that but the ~200 ways I can think of to break TUF, and I would be sure the plain English / reverse engineer method would lead to more mistakes than the JSON-vector-DSL I've proposed.

> description provided... but not in an automated way

That is they beauty of my format: it is (mostly) automated. And it's versioned so that if you want to add some new directive like "incorrectly encode a key" or whatever, you bump the version and mandate that people run a  check like the following to ensure that all the checks have been coded correctly.

```python
PINNED_VERSION = 2

class TestVectors:
    def test_vectors(self, test_vector_meta):
        assert test_vector_meta['version'] == PINNED_VERSION
        for vector in test_vector_meta['vectors']:
            do_test(vector)
```

> Every time some new thing to test occurs to us, we'd need to add some property to the JSON and you'd need to change your parser to support it.

Your proposition is a more manual step than mine because you could add new tests that don't change the version number, and changing a parser in most languages is literally one line. I at least I know it is in Python, Rust, Scala, and Ruby. The other additions (how to change the output) would be equal in both your and my proposed methods.

JustinCappos:
Could we perhaps implement the core of each approach and then compare?  I
see the perspectives from both sides and wonder if a head-to-head is the
best way to resolve this...

awwad:
I think there's still a disconnect. Let's play through your example.

```
  T(kt)  ---k1-->  B(k1) ---k2-->  A(k3)
       \ ------->  C
```

Translating the diagram:
- the Targets role is actually signed by key kt
- the Targets role delegates to B, expecting B to be signed with k1
- B is actually signed by key k1
- B delegates to A, expecting A to be signed by k2
- A is actually signed by key k3

**The TAP proposes that I give you the metadata and a file, keys.json, that tells you what key to sign every role with if you need to convert the metadata.**

For this example, keys.json would look like:

```
{ 'test_repo':  {T: [<kt>], B: [<k1>], A: [<k3>], ...} }
```

**You can have a small tool perform automatic conversion from my format to yours, entirely independently for each role, and do re-signing without having it check for a "sign this with the wrong key" indicator, or any special understanding of the fact that it's dealing with test cases.**

For each role, the metadata converter reads through the 'signed' data, converts it into your format, then signs it with the key indicated by keys.json, whether that is the appropriate key for the role or not.

Does that not work? (Did I miss your point?)

heartsucker:
I think you understood it, but I still think that your format is more limiting. My format describes arbitrary repos and an arbitrary point in time. If you wanted to test some complicated feature, you would have one version a `t0`, another at `t1`, and a final state at `t2`. This might mean set up a repo, "compromise" the targets key, then rotate the root and expect an update to succeed. This test would need three steps, but your format only really allows for 2 (the default trusted state, and single test case).

> Could we perhaps implement the core of each approach and then compare?

@justincappos Can do.

heartsucker:
This is like 90% of the functionality.

https://github.com/heartsucker/tuf-tap-7-example

It takes fixed metadata (I generate it somewhere else) and converts it on the fly into TUF repos.

vladimir-v-diaz:
Thanks for sharing your work on this!  I like the API so far.  One thing that stands out to me ATM is the requirement that clients need to download updates via HTTP.  Why not just make this data available in some archive, directory, etc.?

Also, is this tool meant to used by other implementors?  If so, how do you envision the tool's use if an implementation has other requirements?  After looking at the code, it seems that certain things can be configured, for example: Canonical JSON, standard JSON, key types, etc.

An implementation might use a custom format, which in that case they will have to either modify the tool or convert the tool's generated metadata into its own format.

> If you wanted to test some complicated feature, you would have one version a t0, another at t1, and a final state at t2. This might mean set up a repo, "compromise" the targets key, then rotate the root and expect an update to succeed.

This complicated set up is more of an integration test in my opinion.  We have unit tests in the reference implementation that sort of verify compliance, and then integration tests that attempt to verify the expected behavior that you describe.  I imagine that other implementation would do the same.

I think it might be sufficient to have a test vector that verifies key revocation, which should be doable with one test.

heartsucker:
> Why not just make this data available in some archive, directory, etc?

For my client, once I pass in the ref to a `Repo` object, I can't manipulate the values, so I'd have to so some symlink hacking to get it to point at the next "step." Also, HTTP was easier at first. :)

> Also, is this tool meant to be used by other implementors?

No. I had some of that config stuff left over from another project and just altered this to fit it.

Other implementors would use the same metadata and write their own test harness in their language of choice.

> ... convert the tool's generated metadata into its own format.

Yes, it's possible that someone implementing a conformance tester wouldn't want JSON as the input, in which case it would be trivially easy to convert it to XML or TOML or whatnot.

> This complicated set up is more of an integration test in my opinion.

True, but that's what really matters. Also, I couldn't use my rust lib to generate some of these cases since I don't have access to private members to manipulate values some times, and the logic in the API wouldn't allow me to initialize some structs with incorrect values. Also, because of the way the serializers and deserializers work, they short *could* short cut the initialization logic, so what really matters is "what happens when TUF reads data."

Or, if someone doesn't do HTTP, but MQTT, it would be up to them to implement their conformance test over MQTT instead.

awwad:
#### Single-update vs multi-update
> My format describes arbitrary repos and an arbitrary point in time. If you wanted to test some complicated feature, you would have one version a t0, another at t1, and a final state at t2. This might mean set up a repo, "compromise" the targets key, then rotate the root and expect an update to succeed. This test would need three steps, but your format only really allows for 2 (the default trusted state, and single test case).

I think it's conceptually much simpler to stick to a two-step (single update attempt) process, and it seems to me that it would ask less of an implementer. We don't need these tests to serve as full-scale integration tests (that would constrain implementations too much), and I can't think of an attack test that can't be done in a single update attempt. For a replay attack, for example, one can set the client up to trust a set of metadata that includes version 3 of some role, and then perform an update attempt in which the repo provides version 2 of the same role.

#### Conversion
I didn't see a response to this - did the point on conversion make sense? It's not necessary to reverse engineer or deconstruct or understand what a particular test is trying to do if the metadata can simply be converted. You then won't have to have your own interpreter and test logic for a language of special test instructions like "invalidate xyz signature" or "swap snapshot for previous version of snapshot"; you'd just convert the metadata, which I think is easier to understand and write, and more elegant. (It also means less, if any, disruption for implementers if we add new tests. Instead of interpreting some new command or applying some old command to an unexpected place, you just convert the metadata the same way you previously had - no code changes).

#### Key Index
I thought this was a really good idea, but having thought it through more, I think it breaks what we're doing, and I don't think it provides any special advantage. If we distribute metadata (as, so far, I'm still expecting to), the metadata has to have actual keyids and keys in it. This means that if you need to convert and re-sign and we use key indices in the sign-with-this-key file (keys.json), you would have to map to the key indices in some way so that you'd know what to write in the role files when you're converting metadata and creating your own keys. Distributing actual keys in keys.json would require no more work from us, much less work if the implementer *is* using a compatible key format (and we can distribute test sets using multiple key formats), and no more work from the implementer if they are using incompatible key formats (They still have to adjust values in the metadata the same way when converting.)

#### HTTP
I don't think you're suggesting this, but just in case: TAP 7 itself can't prescribe a distribution mechanism for updates.


After thinking about this, I have a few minor tweaks to make, and then I guess I'll just implement a few tests and test the TUF Reference Implementation with them.

heartsucker:
I still think it's better to write a declarative description of metadata than to provide and convert metadata.

> "swap snapshot for previous version of snapshot"

There's not a special DSL directive for just that case. It's the same as all others:

- at `t0` make snapshot metadata with content X and version 2, use key 1
- at `t1` make snapshot metadata with content Y and version 1, use key 1

The `t0` and `t1` aren't special cases because a test case is already defined as "snapshots of a complete repo at a given time" (what I keep calling a "step"). Content X/Y are roughly equally defined in both our cases. Key 1/`abcdef01234` are roughly equal in both.

This is no more complex than telling someone to parse the version and key ID fields out of the dummy `snapshot.json`. We're just telling them to parse these fields out of the `$test_case.json`.

> I think it's conceptually much simpler to stick to a two-step (single update attempt) process

But TUF *is* complex, and most likely someone who implemented it got all the simple cases right, and it's good to really dive in and reach for as many bizarre edge cases as possible.

>  can't think of an attack test that can't be done in a single update attempt

This may be true now, but why go through all this work only to write a conformance tester generator / parser / converter and not support such cases?

There are test cases that are inexpressible by merely providing pre-existing metadata and asking someone to convert it. A low level DLS in my case is used to construct more complex cases, but this complexity is exposed directly to the implementor as clear directives and not implicitly through metadata parsing and conversion.

In your case, is there a machine readable way to say "create a bad signature?" Because it doesn't seem like it, and if your worry is disruption to developers, my method is minimally invasive one they're written the original implementation. Your method would require an additional DSL on top of already parsing / converting the metadata. Or, you'd have to parse the metadata, canonicalize it, then verify it to see if the sig was bad or not. Right? That is much more complex than `bad_sig: true`.

You partially addressed this above:

> For a conversion case, rather than switching a bit in the signature, we'd probably just provide the wrong key - you wouldn't need to catch that the signature is off, you'd just sign with the key provided, which would be the wrong one.

This is not testing the same logic as a bad signature. If an implementation can't find a key in a trusted set of keys for a role/delegation, it won't even check the signature (most likely). Signing it correctly with the wrong key is not the same as an actual bad signature with the right key. Yours tests business logic about checking that keys match. Mine tests the crypto logic. Both are necessary, but your metadata format can't express all the test cases I think are important for conformance testing.

> you would have to map to the key indices in some way so that you'd know what to write in the role files when you're converting metadata and creating your own keys.

The idea is that the `key_index` is a global value shared by the entire repo, so anywhere you'd have a `key_id`, you just have `key_index`, and map that back to your key store somewhere. You can see that logic here:

https://github.com/heartsucker/tuf-tap-7-example/blob/303aea19e5648311b36441c632f864d81590fd88/tuf_vectors/__init__.py#L238-L252

In both cases, we're doing a look up, but I'm using an int and you're using a string. You even admitted that your scheme only really works if the implementer is using keys that match what you provided. Why bother providing keys? Just tell them to make their own.

> I don't think you're suggesting this

Correct. I just wrote something that generates a TUF repo based of the tuf-vector-json files in the `/metadata` directory. I wrote this to serve them over HTTP so you could play around with it.

vladimir-v-diaz:
> There are test cases that are inexpressible by merely providing pre-existing metadata and asking someone to convert it.

I am having trouble thinking of a test case that couldn't be expressed with pre-existing metadata (apart from a slow retrieval attack).  If one needs to capture an integration test, that should also be possible with pre-existing metadata in my opinion.

@heartsucker 
Can you please share one of these inexpressible test cases?

heartsucker:
The above example: I want a valid key with an invalid signature. I can't express that with @awwad's format.

vladimir-v-diaz:
So this issue arises when metadata is re-signed with new keys and a different format is used?  Yes, this appears to be an issue because the newly-signed metadata would be rendered valid, when in reality we wish to test for a bad signature.

I think in this case I'd recommend we provide a script that should be executed after metadata is re-signed, which can perform any needed edits to the new metadata.  For instance, it can change a bit in "signed" metadata to invalid the signature, etc.

heartsucker:
> we provide a script that should be executed after metadata is re-signed, which can perform any needed edits to the new metadata

That is, again, more complicated than letting an implementer flip the bits themselves. In the simple case where they use JSON but have different field names, your script wouldn't work. In a complex case where they use a binary format, your script would need huge changes. In very complex case where they have a checksum, there is no way the script could handle that.

The number of edge cases that arise in the scenario of parsing/converting pregenerated metadata is much greater than the number of cases with a DSL that describes how to generate the metadata.

vladimir-v-diaz:
Ahh.  Yes, the script wouldn't have a reliable way of changing bits for an unknown format.

PR #61, opened 2017-08-22
Reject TAP 7

JustinCappos:
This rejects TAP 7.


awwad:
> I think @awwad is working on edits to the proposal: #55 
I was, yes. We may as well merge that, too, I think, to reflect the most recent state of the TAP.

vladimir-v-diaz:
Okay, I merged #55.  Feel free to directly edit TAP 7 from now on.

PR #64, opened 2017-08-23
Recommended changes to TAP 4

jhdalek55:
There were a number of comments on the Google Doc. I did not want to
embed them into the text, so I am going to send them either via e-mail
or as a comment on the pull request

### DIFF
--- a/tap4.md
+++ b/tap4.md
@@ -13,7 +13,7 @@
 This TAP offer clients guidance in conducting a secure search for particular
 targets across multiple repositories. It discusses how multiple repositories with
 separate roots of trust can be required to sign off on the same target(s),
-effectively creating an AND relation and ensure the file obtained can be trusted.
+effectively creating an AND relation and ensuring any files obtained can be trusted.
 In other words, this TAP demonstrates how target names can be mapped to
 repositories in a manner similar to the way targets with specific names
 can be delegated to different
@@ -38,8 +38,8 @@ instance (e.g., copy of ```pip```) for each repository.
 ## Use case 2: hiding sensitive metadata and targets
 
 Extending the previous example, enterprise users may not wish to upload some
-metadata and targets to a public repository, because doing so may reveal
-information that is sensitive or proprietary.  Therefore, these
+metadata and targets to a public repository because doing so may reveal
+information that is sensitive or proprietary. Therefore, these
 users may host the sensitive metadata and targets on a private repository,
 while still employing a public
 repository for other files. However, in order to use both private and public
@@ -74,13 +74,13 @@ is already possible to do something similar.  One could have one repository
 download and use a multi-role delegation to the other repository's target.
 Thus, if repository A downloaded the targets metadata from repository B, and
 used a multi-role delegation for the targets metadata, it would achieve a similar
-result.  If repository B is compromised, the users are not impacted because
+result.  In this instance,if repository B is compromised, users are not impacted because
 repository A's multi-role delegation will prevent the use of repository B's
 malicious targets files.  Unfortunately, if repository A's root role or its top level
 targets role were to be compromised, nothing could prevent users from receiving
 malicious targets
 files, even if repository B is not compromised. If, though, adopters required
-valid signatures on the metadata from two repositories, even the compromise described
+valid signatures on the metadata from both repositories, even the compromise described
 above would not impact users.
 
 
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
I prefer we add the Google Docs comments as in-line comments in the pull request.

jhdalek55:
Comments that are more than wording changes that I could not indicate in the text--

_The first paragraph under "specification:_" Is the purpose of this TAP  just  FINDING a file on specific repo, or is it about securely verifying and downloading files from two or more repos? If its just about  looking for a file, is there a risk?--and what then why then would you need multiple signatures if you are not dowloading the file? Isn't the idea of this to make sure the file has not been tampered with? I think this paragraph really needs to take in the broader purpose of what this TAP wants to do.

Also the first two sentences back to back in this paragraph do not follow each other. How does finding a target lead to a compromise? It seems like a sentence is needed between these two thoughts.

The 4th paragraph under "The Map File" subhead...Why are there three bullet points if there are only two keys named?

In the first bullet--what does "target paths of patterns" mean. What are the patterns you are referring to here?

Under "Backwards Compatilbility"--it says "however older clients will not utilize the map file and so will ignore this TAP." If we are arguing that you don't need to use the MAP file to secure multiple repos, then don't you want to make the case that a  client can still perform these actions with another file system?

jhdalek55:
I guess I can re-do it that way if you want.

Lois

vladimir-v-diaz:
It's okay.  I can work with these comments here.

jhdalek55:
Thanks, Vlad.

vladimir-v-diaz:
> In the first bullet--what does "target paths of patterns" mean. What are the patterns you are referring to here?

It is the filename paths listed under the "paths" entry in the map file.  Here is the example used in the TAP:

> // Map any target matching *Django* to both Django and PyPI.
>       "paths":        [`*django*`],
>       "repositories": ["Django", "PyPI"],

`*django*` is an example of a filename with wildcard characters.  `users/lois/tap*.md` is another example. 
 This style of filename path is referred to as a glob pattern.  The [Wikipedia page on Glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)) contains more useful information.

vladimir-v-diaz:
> The first paragraph under "specification:" Is the purpose of this TAP just FINDING a file on specific repo, or is it about securely verifying and downloading files from two or more repos? If its just about looking for a file, is there a risk?--and what then why then would you need multiple signatures if you are not dowloading the file? Isn't the idea of this to make sure the file has not been tampered with? I think this paragraph really needs to take in the broader purpose of what this TAP wants to do.

Clients do eventually download the target file from a single repository, but they must also download metadata about the target from one or more repositories to ensure that there is a repository consensus on the entrusted target.  The number of repositories that must be queried to download the same metadata about a target file depends on the rules specified in the map file (or the equivalent).

vladimir-v-diaz:
I will try to directly address your remaining comments in tap4.md

jhdalek55:
After speaking to Justin yesterday, I was going to make some additional edits this morning. I see the branch has not been merged, so should I make these changes on the same branch? Or start a new one?

Lois

vladimir-v-diaz:
Yup, additional edits to the text should be made to the same branch.  I will see any follow-on edits on this pull request.

jhdalek55:
Cool. So when I finish my changes, I just commit? I don't need to do another pull request. Great. Thanks.

Lois

vladimir-v-diaz:
> So when I finish my changes, I just commit? I don't need to do another pull request.

Yes.  Any new commits will should show up here:
https://github.com/theupdateframework/taps/pull/64/commits

vladimir-v-diaz:
> The 4th paragraph under "The Map File" subhead...Why are there three bullet points if there are only two keys named?

The map file contains two top-level dict keys, named "repositories" and "mapping".  The bullets points you refer to are specifically for the "mapping" key, which can contain the "paths", "repositories", and "terminating" (optionally) keys.  The bullet points cover all of the potential keys that "mapping" can contain, but the "terminating" is optional.  Notice that the last entry in the example only contains  the "paths" and "repositories" keys.  I think this is the reason why the text says, "at least two keys."

I quote the text below:

> The value of the "mapping" key is a priority-ordered list that maps paths (i.e., target names) to the specific repositories. Every entry in this list is a dictionary with at least **two** keys:

vladimir-v-diaz:
Note: I made a commit to your pull request to edit the title.  Notice how it appears under the "commits" tab.
https://github.com/theupdateframework/taps/pull/64/commits

You will need to re-sync with the branch/pull request to make any new commits of your own.

jhdalek55:
I have been working on the file and have made a number of changes. If I
sync up before I commit, will I lose the changes i'm made in the file?

vladimir-v-diaz:
No, Git will attempt to automatically merge your changes with mine. If there are merge conflicts, it will let you know.  However, you shouldn't get any merge conflicts because I have modified a part of the file that you unlikely modified as well.

I recommend that you first commit your local changes and sync after.

Issue #32, opened 2017-05-31
Discussion of TAP 3: Multi-role delegations


vladimir-v-diaz:
TAP 3 proposal is available at https://github.com/theupdateframework/taps/blob/master/tap3.md

Pull requests related to TAP 3:
https://github.com/theupdateframework/taps/pull/3
https://github.com/theupdateframework/taps/pull/7
https://github.com/theupdateframework/taps/pull/9
https://github.com/theupdateframework/taps/pull/11
https://github.com/theupdateframework/taps/pull/12

vladimir-v-diaz:
@JustinCappos @trishankkarthik @awwad 

I came across a few issues with the format of delegations (as proposed in TAP 3) while working on its implementation.  My proposed changes to TAP 3 can be viewed here:
https://gist.github.com/vladimir-v-diaz/827d1513c75c48d70591ee47ed33f38e

Let me know if you agree with the changes and my justification for changing TAP 3.

**Here is a list of the proposed changes:**

(1) "keys" is moved outside of "delegations" and renamed "keys_for_delegations"

(2) Add "name" to "delegations", which is associated with a group/multi-role delegation

(3) Rename "names" to "roleinfo"

(4) "roleinfo" (previously "names") is a list so that delegations are ordered

(5) "threshold" for a multi-role delegation is renamed "min_roles_in_agreement" 


**Reasons for changes:**

(1) If "keys" is kept within "delegations", we'll have to use a dict.  Unfortunately, we lose the ability to order the delegations.  I think "keys_for_delegations" is a clearer name.

(2) In the implementation, I need some way of referring to a muti-role
delegation.  For example, let's say I delegate like so: targets.delegate([role1, role2, ...], ...).  How do I refer to this multi-role delegation?  And how do I revoke it?  And if it's a long list of multiple roles?
I'd like to instead do the following:
repository.targets("claimed").revoke("my_multi-role_delegation")

(3) "names" doesn't quite capture what it contains, and it clashes with "name"

(4) Made into a list so that the roles can be ordered.

(5) Renamed to "min_roles_in_agreement" so that is doesn't clash with
the "threshold" listed inside "roleinfo"

JustinCappos:
So effectively, suggestion 1 just removes the 'delegations' dictionary and promotes / renames the 'roles' item that was within that dictionary.  I don't have a problem with this.  It just is a renaming change, as far as I see.

[...]

PR #73, opened 2017-09-13
Revise TAP 3 (Multi-role delegations)

vladimir-v-diaz:
We decided to amend [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md) to include new metadata fields (e.g., min_roles_in_agreement), convert the "names" field to a list so that delegations can be ordered, etc.  Please visit the [issue discussion for TAP 3](https://github.com/theupdateframework/taps/issues/32) for the full discussion of the proposed revisions to TAP 3.

@awwad Can you please review the revisions and verify that the text is clear?

### DIFF
--- a/tap3.md
+++ b/tap3.md
@@ -195,49 +195,56 @@ targets:
         // This is the first delegation.
         {
           // These targets must be signed by this <b>single</b> role.
+          <b>"name": "first_delegation",</b>
           "paths": ["/foo/*.pkg"],
-          <b>"names"</b>: <b>{</b>
+          "terminating": false,
+          <b>"min_roles_in_agreement": 1,</b>
+          <b>"roleinfo": [
+            {
             // This role must sign them using all 3 of these keys.
-            <b>"alice"</b>: <b>{</b>
+              "rolename": "alice",
               "keyids": [
                 "ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb",
                 "acac86c0e609ca906f632b0e2dacccb2b77d22b0621f20ebece1a4835b93f6f0",
                 "de7d1b721a1e0632b7cf04edf5032c8ecffa9f9a08492152b926f1a5a7e765d7"
               ],
               "threshold": 3
-            <b>}</b>
-          <b>}</b>,
-          ...
+            }</b>
+          ]
         },
         // This is the second delegation.
         {
           // These targets must be signed by <b>both</b> of these roles.
+          <b>"name": "second_delegation",</b>
           "paths": ["/baz/*.pkg"],
           "terminating", true,
-          <b>"names"</b>: <b>{</b>
+          <b>"min_roles_in_agreement": 2,</b>
+          <b>"roleinfo": [
             // The release engineering role must sign using this key.
-            <b>"release-engineering"</b>: <b>{</b>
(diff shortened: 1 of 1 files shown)


tap3.md
```diff
@@ -96,68 +94,88 @@ the given set of targets.
 }
 </pre>
 
-## The new targets metadata file format
+## The new file format of targets metadata
```
awwad: You provide two files in the new format:
- a generic format definition with some examples in it
- a better example

I think this is confusing in a reference document and also generally to a casual reader. I think the format definition should be entirely generic, with one delegation listed and a snipped second (just to indicate that there's a list of delegations), and no specific values, just labels. Then you can move on to the better example you have below.

I can PR or commit that if you don't mind.

You may even want to exclude most of the comments in the definition and keep them in the example.

tap3.md
```diff
@@ -174,54 +192,69 @@ targets:
...
         {
           // These targets must be signed by this <b>single</b> role.
+          <b>"name": "first_delegation",</b>
           "paths": ["/foo/*.pkg"],
-          <b>"names"</b>: <b>{</b>
+          "terminating": false,
+          <b>"min_roles_in_agreement": 1,</b>
+          "roleinfo": [
+            {
             // This role must sign them using all 3 of these keys.
```
awwad: I suggest cutting this line and adding some description 8 lines up, in the delegation description, explaining the full meaning of this delegation there: "This first delegation requires one role ("min_roles_in_agreement" = 1), "alice", to sign targets matching pattern "/foo/*.pkg". It also specifies that the "alice" role must be signed with all three of the keyids listed ("threshold" = 3) in order for to be validated."


I guess you could alternatively move this comment line down three lines into "keyids" and write instead:
'These three keys are being associated with this role. Per "threshold", all three must be used to sign this role. Per "min_roles_in_agreement" and "roleinfo", only this role, "alice", need sign files matching /foo/*.pkg in order for those files to be validated.'

I prefer the first.

tap3.md
```diff
@@ -174,54 +192,69 @@ targets:
...
                 "de7d1b721a1e0632b7cf04edf5032c8ecffa9f9a08492152b926f1a5a7e765d7"
               ],
               "threshold": 3
-            <b>}</b>
-          <b>}</b>,
-          ...
+            }
+          ]
         },
         // This is the second delegation.
```
awwad: I suggest you crop some of the comments below and write this here:

This, second, delegation requires that both roles ("min_roles_in_agreement" = 2) 'release-engineering' and 'quality-assurance' sign for the target files in order to validate them via this delegation. 'release-engineering' is expected to be signed with the single listed key for that role ("threshold" = 1). 'quality-assurance' must be signed with at least two of the three listed keys for that role ("threshold" = 2).

tap3.md
```diff
@@ -86,7 +86,7 @@ sign the given set of targets.
           "terminating": false,
         }
         // Note that, unfortunately, there is no way to require <b>multiple</b>
-        // roles to sign targets in a single delegation.
+        // roles as required to sign targets in a single delegation.
```
awwad: this one is redundant now. "require multiple roles as required"

don't need the change

vladimir-v-diaz:
Thanks for the review, Seb.

PR #76, opened 2017-09-20
Refining text

jhdalek55:
This addresses the two phrases questioned in Justin’s last review.

I checked for and found a few other minor wording issues as well

Justin's original review:

> It's not clear why there is an ordered list of repositories when you talk about: "A list of target paths associated with each ordered list of repositories. "  Why more than one?

> "The exact design of this mechanism is left to adopters, but, at a minimum, it must support or exhibit the following three properties:" The first part of this is odd.  What do you mean to say?


vladimir-v-diaz:
Merging!

I edited the pull request summary to add Justin's review.

PR #77, opened 2017-09-26
revise tap8

hannesm:
this is a major revision of TAP8.  please let me know what you think.  I'm sorry that I may have missed some feedback.

### DIFF
--- a/tap8.md
+++ b/tap8.md
@@ -1,7 +1,7 @@
 * TAP: 8
 * Title: Key rotation and explicit self-revocation
 * Version: 1
-* Last-Modified: 10-May-2017
+* Last-Modified: 26-Sep-2017
 * Author: Hannes Mehnert, Justin Cappos
 * Status: Draft
 * Content-Type: text/markdown
@@ -12,20 +12,21 @@
 
 TAP 8 generalises the mechanism of chaining public key generations, and
 provides a secure way for key rotation.  Rotation is the process by
-which a*role uses their old key to invalidate their old key and transfer
+which a role uses their old key to invalidate their old key and transfer
 trust in the old key to a new key.  Performing a key rotation does not
 require parties that delegated trust to the old key to change their
 delegation to the new key.  Conceptually, the rotation process says if
 you trusted key X (or threshold of keys X_0, ... X_n), now instead trust
 key Y (or threshold of keys Y_0, ... Y_n).  Rotation of a key may be
-performed any number of times, transferring trust from Y to Z, etc.
+performed any number of times, transferring trust from X to Y, then from
+Y to Z, etc.
 
 The mechanism in this TAP has an additional use case:  if a rotation
 cycle (A to B, B to A) is detected, all delegations into the cycle are
 invalid.  Note, this is handled the same as a delegation to a missing key
 and does not invalidate the delegator's entire file.  This use of rotate
-loops allows a role to explicitly revoke their own key for all future 
-actions (by introducing a cycle).  
+loops allows a role to explicitly revoke their own key for all future
+actions (by introducing a cycle).
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


tap8.md
```diff
@@ -93,48 +111,115 @@ introduced, which contains the new public key(s), the threshold, and is
 signed by a threshold of old public keys.   The intuition is while
```
vladimir-v-diaz: Consider omitting `file` in the first sentence, so that the sentence instead reads: "To support key rotation, the new metadata type `rotate` is introduced..."  I thought `rotate` was the file name when reading it the first time.

tap8.md
```diff
@@ -93,48 +111,115 @@ introduced, which contains the new public key(s), the threshold, and is
...
-signatures on foo.rotate.hash(ABC2).  The new targets file,
-foo.hash(ABCD2) is verified using the keys ABCD from the rotate file.
-
-When Evelyn joins, and the threshold is increased to 3, foo.hash(ABCD2)
-is replaced with the rotate file foo.rotate.hash(ABCD2) containing ABCDE
-keys and the new threshold of 3, signed by two of ABCD.  Delegations can
-point to either foo.hash(ABC2), foo.hash(ABCD2), or foo.hash(ABCDE3),
-they have all the same meaning.
+file to be foo.  When they want to add Dan to the project, they create
+a foo.rotate.Alice.Bob.Charlie.2 file, which contains all four keys, and a
```
vladimir-v-diaz: In TAP 3, a multi-role delegation is given a single name, and only individual keyids are listed.  In other words, individual names like `Alice`, `Bob`, etc. are associated with keyids.

See the [multi-role example in TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md#example-requiring-a-combination-of-roles-to-sign-the-same-targets).

tap8.md
```diff
@@ -93,48 +111,115 @@ introduced, which contains the new public key(s), the threshold, and is
...
+targets file using the new key.  If the new keyid is still not used in
+the signatures, another rotate file needs to be downloaded,
+`foo.rotate.NEW_KEYID.1`, until either a valid chain is found, in which case
+the targets file is valid, or key data is missing or there is a cycle in
+the rotations, in both cases the targets file is invalid.
+
+## Root rotation
+
+Root, timestamp and snapshot key rotation.  These keys can rotate as
+well, leading to ROLE..rotate.HHH.T files, where HHH is the concatenation of
```
vladimir-v-diaz: Can't these filenames get really long if there are many concatenated keyids?

tap8.md
```diff
@@ -93,48 +111,115 @@ introduced, which contains the new public key(s), the threshold, and is
...
+
+
+## Client workflow
+
+A client who wants to install foo now fetches Alice targets file, and
+fails to verify the signature with Alice old key - since none of the
+keyids in the signatures match it.  The client fetches
+`foo.rotate.KEYID.THRESHOLD`, where KEYID is the key identifier the
+client found in the delegation to Alice, and THRESHOLD the old threshold, as
+ASCII encoded integer (0x31 for 1).  The client can verify this rotate
```
maxyunrock: Need a little clarification about implementation of rotate:
The filename suffix, referred as ID above and below, is the hex representation of the SHA256 hash of the concatenation (using "." as separator) of the KEYIDs (in ascending lexical order), and the old threshold value, encoded decimal as ASCII (0x31 for 1, 0x32 for 2, 0x31 0x30 for 10).

The following have been interpreted: 
hex{sha256(KeyID.1)}
hex{sha256(KeyID.0x31)}
hex{sha256(KeyID.31)}
which one is intended?
Could it be a issue if filenames get long ? Any alternatives?

Goldgaruda:
There are folds named metadata and metadata.staged under the repository.
What is the differences between the two folds.
I mean what the meaning is respectively. Some json files exist under both folds.

Goldgaruda:
In TAP8, the client verifies this rotate file using the public key from the delegation. 

Where does the delegation come from?   Json file? I noticed that only targets.json has delegation parts.
So the key is in delegation from the corresponding targets.json file?

Goldgaruda:
if a rotation cycle (A to B, B to A) is detected, all delegations into the cycle are invalid.  
Here, regarding A and B, what does they mean? Is it a  key cycle (i.e.,  X_0, ... X_n-->Y_0, ... Y_n....--> X_0, ... X_n) or a role cycle (Alice--> Bob---> Christina..... --->Alice)?

JustinCappos:
Cycles of delegations are allowed (and fairly common).  It is key rotation
cycles that then fail because all keys are considered revoked.

Justin

Goldgaruda:
How to maintain a key table/record file for those keys used.

If keys are rotated, how can we  check key rotation efficiently?

Problem:
If we detect the cycle of a key rotation, how to deal with old keys. 
Just delete it? or move them to a special repository or just mark them as used labels .
Considering the number of cycles, old keys, is there any efficient alterative ?

JustinCappos:
There are separate files for each rotation according to TAP 8, right?   Why
do you need a table?

Justin

Goldgaruda:
We think about that from the viewpoint of efficiency. 
Since the rotation files are immutable and should not be replaced or modified under any circumstances, there should be lots of  rotation files under the meta directory, which potentially cause the efficiency problems when doing the circle check.

JustinCappos:
I believe one can remove them if there are only delegations to later keys.
So at least clients need not download them in this case.  Really, each
client downloads each key once (at most).  Since the metadata files are
fairly small, this may be reasonable overhead.  I guess it depends on the
rate of key rotation...

I don't think efficiency of checking rotation files is a concern.  What
algorithm were you planning to use to check?

tap8.md
Goldgaruda: -----------------------------------------------------------------------------------
Let's consider a motivating example, project foo is delegated to Alice. Alice computer with the key material got stolen, but the disk was encrypted. To be safe, she decides to get her key from her backup and roll over her key to a fresh one. To do that, she creates a file foo.rotate.ID.
-----------------------------------------------------------------------------------------------------
For foo.rotate.ID, in TAP 6, foo is said to be the project name.
When generating the rotate file, where does the value of project name come from?

Goldgaruda:
Hi professor,
Because we need the old key information to locate the corresponding rotate file whose name (projectname.rotate.ID) is encoded using SHA256. If we don't store the former keys, we cann't check whether the keys are used or not before.

In addition, for the key rotation cycle check (i.e., in the statement if a rotation cycle (A to B, B to A) is detected), is such cycle only limited in 2 key chain, I mean, just for A->B->A?  or it can be a very long cycle, e.g., A->B->C....X->Y->Z-A.

PR #78, opened 2017-09-29
address @vladimir-v-diaz comments

hannesm:
importantly, use as suffix the SHA256 of the concatenation (with `.` as separator) of KEYIDS and threshold.

thanks @vladimir-v-diaz for your comments -- I hope this PR makes it clearer!

### DIFF
--- a/tap8.md
+++ b/tap8.md
@@ -1,7 +1,7 @@
 * TAP: 8
 * Title: Key rotation and explicit self-revocation
 * Version: 1
-* Last-Modified: 26-Sep-2017
+* Last-Modified: 29-Sep-2017
 * Author: Hannes Mehnert, Justin Cappos
 * Status: Draft
 * Content-Type: text/markdown
@@ -10,8 +10,7 @@
 
 # Abstract
 
-TAP 8 generalises the mechanism of chaining public key generations, and
-provides a secure way for key rotation.  Rotation is the process by
+TAP 8 generalises the mechanism of key rotation.  Rotation is the process by
 which a role uses their old key to invalidate their old key and transfer
 trust in the old key to a new key.  Performing a key rotation does not
 require parties that delegated trust to the old key to change their
@@ -30,8 +29,8 @@ actions (by introducing a cycle).
 
 # Motivation
 
-TUF supports key rotation using multiple ways that are ad hoc and differ
-based upon the mechanism.  This leads to some concerns:
+TUF supports key rotation for the root keys using an ad hoc way.
+Several use cases can benefit from a generalised mechanism:
 
 
 ## Self-maintaining *project* role
@@ -71,7 +70,7 @@ old ones are removed need to be fetched and verified.
 Roles which typically have short lived keys, like the timestamp role,
 may wish to revoke trust in the prior key and sign a new key with the
 old.  This limits the ability for attackers to walk back the set of
-trusted keys.  Right now, there is not a good way to do this within TUF,
+trusted keys.  Right now, there is no good way to do this within TUF,
 which may result in some keys being trusted more than they should be.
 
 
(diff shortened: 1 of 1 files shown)


tap8.md
```diff
@@ -130,94 +129,94 @@ signatures part as in tuf spec, not shown here):
...
-her targets file with the new key, and uploads both the rotate file
-and the freshly signed targets file to the repository.
+`foo.rotate.ID`.  This file contains her new key, a threshold of 1, and
+signed with her old key.  She signs her targets file with the new key,
+and uploads both the rotate file and the freshly signed targets file to
+the repository.
 
+The filename suffix, refered as `ID` above and below, is the hex
+representation of the SHA256 hash of the concatenation (using "." as
+separator) of the KEYIDs (in ascending lexical order), and the old
```
vladimir-v-diaz: It's not clear to me which KEYIDs are used in this case, the old or the new KEYIDs?  I recommend that you elaborate at this point in the text. 

What adds to my confusion are the following statements that appear to say different things.  Line 158 states, "ID is explained above using Alice old keyid."  Later in the text,  on line 163, it says, "where ID' is constructed using Alice's new keyid"

tap8.md
```diff
@@ -130,94 +129,94 @@ signatures part as in tuf spec, not shown here):
...
-public keys, separated by a ".".  The value
-of T is the ASCII encoded old threshold value.  Each file is signed by
-a quorum of keys HHH, and contains the new keys.  A client can fetch the
-actual data, timestamp, and verify it.  If the keyids do not match, the
-client needs to fetch the ROLE.rotate.HH file
-where HH is the keyid of the last trusted timestamp key, either from the
-root file or locally cached.  If the timestamp key is renewed by the
-root, all timestamp.rotate files can be safely removed from the
+well, leading to ROLE.rotate.ID files, where ID is as described above.
+The value of T is the ASCII encoded old threshold value.  Each file is
```
vladimir-v-diaz: What is `T`?

tap8.md
```diff
@@ -130,94 +129,94 @@ signatures part as in tuf spec, not shown here):
...
-threshold (of 2) to Alice, Bob, and Charlie.  Let's assume the targets
-file to be foo.  When they want to add Dan to the project, they create
-a foo.rotate.Alice.Bob.Charlie.2 file, which contains all four keys, and a
-new threshold (again 2).   The file foo.rotate.Alice.Bob.Charlie is signed
-by at least 2 of Alice, Bob, and Charlie.  The new targets file foo
-is then signed by a new threshold (again 2) of Alice, Bob, Charlie, and
-Dan to complete the rotation.
+threshold (of 2) to keyids Alice, Bob, and Charlie.  When they want
+to add a keyid from Dan to the project, they create a foo.rotate.ID
+file, where ID is as described above (the SHA256 of the concatenated
```
vladimir-v-diaz: Should you instead say, "the hex representation of the SHA256 ..." ?

tap8.md
```diff
@@ -130,94 +129,94 @@ signatures part as in tuf spec, not shown here):
...
 tuf spec.  The value of ROLE has to be the same as the role for the
-delegation.  The value for THRESHOLD is its new value.  The keys
+delegation.  The value of THRESHOLD is its new value.  The keyids
 specify the new valid key ids (which may be a subset or superset of
-the old ones).  The signatures have to be done with the old keys.
-There is no expiration of a rotate file, it is not meant to be
-ever modified or in need of re-signing.
+the old ones).  A rotate file does _not_ contain an expiration date,
+it is meant to be signed once and never modified.  The rotate
+file has to be signed with an old threshold of old keys.
```
vladimir-v-diaz: I wonder if we should explain that rotate files are never modified because users/roles are expected to upload new rotate files to the repository, and that these rotate files are expected to have unique filenames (as I understand).  This is due to the new `ID` value in `foo.rotate.ID`, to use an example.

PR #89, opened 2017-11-03
Remove Use Case 3, which is covered by 1 and 4 (now 3)

### LINKED ISSUE #88: Use Case 3 in TAP 4 doesn't seem to make sense
Use Cases [1](https://github.com/theupdateframework/taps/blob/006fc7ca520f7cad0c3228e24b07f9b1de48cd58/tap4.md#use-case-1-obtaining-different-targets-from-different-repositories) and [4](https://github.com/theupdateframework/taps/blob/006fc7ca520f7cad0c3228e24b07f9b1de48cd58/tap4.md#use-case-4-improving-compromise-resilience-given-multiple-repositories) cover clients selecting which repository to trust for which targets and trusting a combination of multiple repositories for a given target, and so I think they already cover what [Use Case 3](https://github.com/theupdateframework/taps/blob/006fc7ca520f7cad0c3228e24b07f9b1de48cd58/tap4.md#use-case-3-verifying-the-same-target-against-different-repositories) intends to demonstrate.

I think Use Case 3 just confuses things. If it's about client choice, that's covered in 1 and 4, and if it's about repository maintainer choice, then that's actually provided by TAP 3, counter to what Use Case 3 suggests.

awwad:
Fixes #88. Please see that issue.

### DIFF
--- a/tap4.md
+++ b/tap4.md
@@ -44,19 +44,7 @@ private and public repositories, TUF clients need to know how to search for
 selected targets on the private repository, and all other targets on the public
 repository.
 
-## Use case 3: verifying the same target against different repositories
-
-[TAP 3](tap3.md) allows metadata publishers to require multiple roles to sign
-off on targets, which forces clients to verify against all roles listed by the
-metadata. While this is desirable in some instances, other clients may prefer
-to require this step only for certain roles. For example, a [Continuous
-Integration](https://en.wikipedia.org/wiki/Continuous_integration) system may
-only care about developer signatures, while a staging environment may wish to
-verify both developer and CI signatures. In order to support this selective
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
Please update the TAP's `Last-Modified`.

JustinCappos:
Done.  Should that have been part of the original PR or do we do it when
merging?

vladimir-v-diaz:
We don't have an official guideline for this.  I often try to bump the `Last-Modified` entry in the original pull request.  However, it can require less work to do it with the merge.  For instance, the original pull request might need edits days later after undergoing review, so the original bump is a waste.  It's probably better to hold off on bumping `Last-Modified` until the merge.

PR #100, opened 2018-01-16
Tap 3: Remove unneeded level of new format, 'roles' in 'delegations'

awwad:
This seems to have been left in by mistake.

This PR makes the 'delegations' element a list, each
element of which is a dictionary containing the data for a
single delegation.

Previously, instead, 'delegations' was a dictionary always containing a
single item called 'roles', which was the list described above.

Aside from being unnecessary, 'roles' was also a misleading
name for a list of delegations.

### DIFF
--- a/tap3.md
+++ b/tap3.md
@@ -116,8 +116,7 @@ _multiple_ role names as required to sign, instead of a single one.
 {
   "signed": {
     <b>"keys_for_delegations": {KEYID: {"keytype: KEYTYPE, "keyval": KEYVAL}, ...},</b>
-    "delegations": {
-      "roles": [
+    "delegations": [
         {
           <b>"name": DELEGATION-NAME,</b>
           "paths": [PATHPATTERN-1, PATHPATTERN-2, ...],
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
I discussed this change with Sebastien and it makes sense.  I'l wait for any potential objections before merging..

vladimir-v-diaz:
Can you verify that all instances of "roles" is removed?  The `new file format of targets metadata` section appears to still specify the "roles" attribute.
https://github.com/awwad/taps/blob/f2941d7f22bd36b6fb64eb7118df7aff8b3a44ec/tap3.md#the-new-file-format-of-targets-metadata

awwad:
Quite right. Fixing the fix, thanks. (:
(Also tidied up the fixed brackets.)

vladimir-v-diaz:
LGTM.

vladimir-v-diaz:
Sorry, one last thing.  Can you update the "Last-modified" field?

vladimir-v-diaz:
Thanks!  Merging..

PR #101, opened 2018-01-22
TAP5: minor improvements

hannesm:
this doesn't change any semantics, just some markdown-formatted improvements (i.e. making the link an actual link to the paper etc.)


vladimir-v-diaz:
Okay.  Thanks for the update.

I updated the `Last-Modified` entry.

PR #103, opened 2018-06-18
Draft of TAP 12 (1.0 release schedule)

vladimir-v-diaz:
This pull request adds a draft of TAP 12 (1.0 release schedule).  It covers the release schedule and anticipated features for the 1.0 release of the reference implementation.

Note: We need to settle on an EOL for the 1.0 release, preferably with community input.

Signed-off-by: Vladimir Diaz <vladimir.v.diaz@gmail.com>

### DIFF
--- a/tap12.md
+++ b/tap12.md
@@ -1,0 +1,84 @@
+* TAP: 12
+* Title: 1.0 release schedule
+* Version: 1
+* Last-Modified: 18-June-2018
+* Author: Vladimir Diaz
+* Status: Draft
+* Type: Informational
+* Content-Type: text/markdown
+* Created: 18-June-2018
+
+# Abstract
+
+This TAP covers the release schedule for the 1.0 release of the reference
+implementation.  1.0 releases match the `1.X.Y` distribution identifier, and
+conform with the official [1.0
+specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md)
+
+Note: The distribution version scheme of the project follows the
+(major.minor.micro) format.  Pre-releases of the project have a fixed `major`
+of `1`, while the `minor` and `micro` segments can vary.  On occassion,
+developmental releases are made available on PyPI for testing and follow the
+`1.X.Y.devZ` version scheme.
+
+# Release manager for 1.0 releases
+Vladimir Diaz <vladimir.v.diaz@gmail.com, @vladimir-v-diaz>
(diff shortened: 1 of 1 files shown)


JustinCappos:
We'll do this, but I don't think the right way is with TAPs.  If you feel
strongly we should use the TAP process for this, I suppose we can discuss,
but I'd rather announce and do this separately.

trishankatdatadog:
That's fine, but we need to formalize the process somewhere, and I can't think of anything better than TAPs right now for this. It seems like a good idea to hold commitments somewhere. IIRC, Python uses PEPs to plan release schedules and backwards-incompatibilities, too.

@SantiagoTorres What are your thoughts?

awwad:
TAPs are intended, I think, more for the TUF specification than for the reference implementation, and release schedules relate to the reference implementation.

Is there an issue with, say, publishing release candidates that indicate when they'll be succeeded by an official release (and of course indicating in the release notes where backward compatibility is threatened)?  That seems more lightweight but also like it would serve your needs?  I'd imagine that you wouldn't necessarily want to have to wait for a scheduled release if something important was missing.  We're also a pretty lightweight operation, so I don't know that we need something so rigid as TAPs for releases.

SantiagoTorres:
Well this is part of what irks me about TAPs in a sense. Are they for the TUF project or are they related to just the specifciation? Quoting TAP 1:

>  A TAP is a design document providing information to the TUF community, or describing a new feature for TUF or its processes or environment.

I don't really see why it couldn't be a process TAP to describe a release schedule. I also don't really know where else could it be discussed that's formal enough. I'd rather avoid having a release schedule put in a github ticket or an email thread.

trishankatdatadog:
@awwad It's not enough to just warn in a changelog. People might be installing TUF from a CI/CD pipeline, and not expect things to break.

@vladimir-v-diaz and I had an informal agreement that as long as you constrained pip to the MAJOR.MINOR.* line (e.g., 0.11.*), you should not see breaking changes, but this needs to be formalized.

trishankatdatadog:
@SantiagoTorres 💯

SantiagoTorres:
> @vladimir-v-diaz and I had an informal agreement that as long as you constrained pip to the MAJOR.MINOR.* line (e.g., 0.11.*), you should not see breaking changes, but this needs to be formalized.

This sounds very reminiscent of semver. I assume formalizing it as "we're adopting semver with these specifics" would suffice?

trishankatdatadog:
@SantiagoTorres Yes, and we need to formalize this somewhere

PR #105, opened 2018-10-15
Tap8revision

mnm678:
Based on discussion in the issue, I made a few proposed changes to the TAP 8 spec. These are:

- The root role does not use the TAP 8 rotation mechanism, but continues to use the ad hoc rotation.
- Rotate files can be replaced by a revocation, but nothing else.
- Each rotate file points to the previous rotate file to prevent accidental rotation cycles. Key revocation is done through a rotation to a null key.
- Repositories contain a rotate folder that holds all rotate files.

### DIFF
--- a/tap8.md
+++ b/tap8.md
@@ -21,35 +21,40 @@ Performing a key rotation does not require parties that delegated trust to the
 role to change their delegation. Roles are thus able to rotate to new trusted
 keys independently, allowing keys to be changed more often.
 
-Conceptually, the rotation process says if you trusted threshold of keys 
-X_0, ... X_n, now instead trust threshold of keys Y_0, ... Y_n.  Rotation of a 
-key may be performed any number of times, transferring trust from X to Y, then 
-from Y to Z, etc. Trust can even be transfered back from Z to X, allowing a key
+Conceptually, the rotation process says if you trusted threshold of keys
+X_0, ... X_n, now instead trust threshold of keys Y_0, ... Y_n.  Rotation of a
+key may be performed any number of times, transferring trust from X to Y, then
+from Y to Z, etc. Trust can even be transferred back from Z to X, allowing a key
 to be added to a role, then later removed.
 
 The mechanism in this TAP has an additional use case: if a rotation
 to a null key is detected, it causes the role to no longer be trusted.  
 A role could use a rotation to a null key if they suspect a compromised key
-(a lost hard drive, system breach, etc). The role is able to create a 
-rotation to null without the help of the delegator, so they are able to 
+(a lost hard drive, system breach, etc). The role is able to create a
+rotation to null without the help of the delegator, so they are able to
 explicitly revoke trust in the role immediately, improving response time
 to a key compromise. A rotation to a null key revokes trust in the role,
 not specific keys, so all keys associated with the role will be invalid
-after a rotation to null. The client will detect a rotation to a null key 
-and treat it as if the delegation was missing a key. The rest of the 
+after a rotation to null. The client will detect a rotation to a null key
+and treat it as if the delegation was missing a key. The rest of the
 delegator's file would still be valid.
 
 The delegator is able to recover from a rotation to null by delegating
 to a new role. The role that has been self revoked can no longer be used.
 Additionally, a delegator can overrule a rotate file by delegating to a
 new role. This ensures that the delegator still has the root of trust,
-but allows the role to act independently.
+but allows the role to act independently. The exception to this is the roles
+defined in the TUF spec (timestamp, snapshot, etc). These roles cannot be
+delegated to a different rolename, so the root role can delegate to the same
+role with a different key to recover from a rotate to null. The root role
+is responsible for ensuring that the current chain of trust is valid
+(this process is described in detail below).
 
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


tap8.md
```diff
@@ -118,116 +140,116 @@ signatures part as in tuf spec, not shown here):
...
 
 Let's consider the project foo, initially delegated to a multi-role
 threshold (of 2) to keyids Alice, Bob, and Charlie.  When they want
-to add a keyid from Dan to the project, they create a foo.rotate.ID
+to add a keyid from Dan to the project, they create a foo.rotate.ID.PREV
 file, where ID is as described above (the SHA256 of the concatenated
-key ids of Alice, Bob, Charlie, and the character 0x32).  This
-contains all four keys, and a new threshold (again 2).   The file
-foo.rotate.ID is signed by at least 2 keyids of Alice, Bob, and Charlie.
+key ids of Alice, Bob, Charlie, and the character 0x32) This
```
hannesm: after `0x32)` a dot (`.`) is missing

hannesm:
thanks! remarks:
- the `null` key should be explicitly listed in my opinion (so that implementations can use this sentinel value in a interoperable way)
- the `null` key is used for (a) self-revocation and (b) for initial rotation (where `PREV` points to SHA256 of `null`). my opinion is that instead using `null` only for (a), and the empty value (`""`) for (b).
- the `null` key is never part of the repository, but instead a hardcoded sentinel in all tuf implementations, is this correct?

tap8.md
```diff
@@ -10,45 +10,73 @@
...
-key may be performed any number of times, transferring trust from X to Y, then 
-from Y to Z, etc.
-
-The mechanism in this TAP has an additional use case:  if a rotation
-cycle (A to B, B to A) is detected, all delegations into the cycle are
-invalid.  Note, this is handled the same as a delegation to a missing key
-and does not invalidate the delegator's entire file.  This use of rotate
-loops allows a role to explicitly revoke their own key for all future
-actions (by introducing a cycle).
+TAP 8 allows roles to rotate trust from their current keys to a new key or keys.
```
JustinCappos:
```suggestion
TAP 8 allows a role to change or revoke their key without requiring changes from
parties that delegate to that role.  This involves a role rotating trust from their 
current key(s) to a new key or keys.
```

tap8.md
```diff
@@ -58,12 +86,11 @@ the initial root is trusted (TOFU or by being distributed over a trusted
 channel), following the chain of key rotations leads to verified keys
```
JustinCappos: You should clarify before this that root rotation does not change for this TAP.

tap8.md
```diff
@@ -58,12 +86,11 @@ the initial root is trusted (TOFU or by being distributed over a trusted
...
-order to follow the chain of delegations.  This could be an efficiency
-concern in some situations.
+order to follow the chain of delegations.  The root metadata contains
+additional information, like the spec-version that needs to be seen for
+each version of the file.
 
-TAP 8 provides a unified methodology for key rotations.  Using this
-mechanism, only those root files in which new root keys are added or
-old ones are removed need to be fetched and verified.
+TAP 8 provides a methodology for key rotations for other roles as well.
```
JustinCappos: What do you mean?  You say this doesn't work for root rotation.  Why is it "other roles as well"?

tap8.md
```diff
@@ -58,12 +86,11 @@ the initial root is trusted (TOFU or by being distributed over a trusted
 channel), following the chain of key rotations leads to verified keys
 for the current root metadata.  This means that even if the keys are not
 changed, a client must download every version of the root metadata in
-order to follow the chain of delegations.  This could be an efficiency
-concern in some situations.
+order to follow the chain of delegations.  The root metadata contains
+additional information, like the spec-version that needs to be seen for
+each version of the file.
```
JustinCappos: I'm not sure why this is relevant.

tap8.md
```diff
@@ -87,26 +113,32 @@ thus is not intended to be handled by this TAP.
...
-TUF is a framework for securing software update systems.  The trust
+TUF is a framework for securing software update systems that is designed
+to maintain trust in the system through key compromises. The trust
 is rooted in a quorum of root keys, which delegate trust to other roles.
 When the root keys delegate trust to a role t for all names, this can be
 used to delegate some names to role d, etc.  When the person who owns
-role d wants to renew their key, they have until now to ask the holder of
-role t to delegate to the new keyset d'.
+role d wants to renew their key, they have until now had to ask the holder of
+role t to delegate to the new keyset d'. If one of role d's keys is
```
JustinCappos: given d's looks possessive and you are trying to use d' as d prime, perhaps different notation is needed.

tap8.md
```diff
@@ -248,17 +290,15 @@ provide a simple mechanism extending and shrinking projects by
 themselves without an individual with elevated privileges, but based
 on a threshold of signatures.
 
-Clients need to take care to check for 'a rotation loop' where key
-rotations point to other keys in a cycle.  This should be handled in the
+Clients need to take care to check for null rotations (rotate
```
JustinCappos: is "rotation to null key" less confusing?  Why use "null rotations"?

hannesm:
I re-read this PR and it looks good to me! thanks!

PR #107, opened 2018-12-20
Added TAP for TUF Version Management

mnm678:
This TAP clarifies how breaking changes can be implemented in TUF and ensures that clients are aware of these changes.


tap12.md
```diff
@@ -0,0 +1,54 @@
...
+
+# Specification
+
+The root metadata already contains the TUF spec-version. The client shall compare the spec-version in the root metadata (server spec-version) with the spec-version of the local client (client spec-version). The client shall then proceed as described below.
+
+## Procedure
+
+If the server spec-version is lower than the client spec-version, the client shall terminate the update and report an error.
+
+If the major version (the first digit) of the spec-version has been incremented, the client must update before proceeding. This could be an automatic process or an error could be reported, requesting a manual client update.
```
trishankatdatadog: Thanks for the TAP, Marina!

Not clear what this means, though. The client must update what from where?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,142 @@
...
+* Version: 1
+* Last-Modified: 22-July-2019
+* Author: Marina Moore, Justin Cappos
+* Status: Draft
+* Content-Type: text/markdown
+* Created: 19-December-2018
+
+# Abstract
+
+This TAP clarifies how to manage updates to the TUF specification that include non-backwards compatible (breaking) changes. A breaking change means that a client and repository must implement the change at the same time in order to continue functioning as expected. This includes any feature that affects the parsing or validation of metadata, or any other changes that mean a client can no longer safely and reliable perform an update if the repository is updated out of sync with the client. To facilitate finding breaking changes, this TAP defines a procedure for TUF clients to ensure that their version of the TUF specification is compatible with the specification version of the metadata they download. This TAP allows clients and repositories to update from their current version of the TUF specification to a version that is not backwards compatible.
```
JustinCappos:
```suggestion
This TAP clarifies how to manage updates to the TUF specification that include non-backwards compatible (breaking) changes. A breaking change means that a client and repository must implement the change at the same time in order to continue functioning as expected. This includes any feature that affects the parsing or validation of metadata, or any other changes that mean a client can no longer safely and reliably perform an update if the repository is updated out of sync with the client. To facilitate finding breaking changes, this TAP defines a procedure for TUF clients to ensure that their version of the TUF specification is compatible with the specification version of the metadata they download. This TAP allows clients and repositories to update from their current version of the TUF specification to a version that is not backwards compatible.
```

Is the TUF style to line wrap TAPs?  I'm finding it slightly painful to suggest edits in this format...

JustinCappos: > This TAP allows clients and repositories to update from their current version of the TUF specification to a version that is not backwards compatible.

This phrasing is really odd.  It may be easier to explain what we are intending to support.  I assume you'll explain how this TAP mandates that repos generate metadata for these cases and how clients can update to a version that supports the right metadata version.  You will need to explain also what assumptions you are making (the server will always generate duplicate copies of old metadata, the client updater will always be able to update to support the latest version of TUF, etc.)

candidate-tuf-versions.md
```diff
@@ -0,0 +1,142 @@
...
+
+As they do not affect compatibility, non-breaking changes may happen independently on clients and repositories. Any specification versions that do not contain breaking changes may be updated to at any time without worrying about compatibility. This TAP clarifies how TUF version numbers are formatted so that breaking changes are distinct from non-breaking changes.
+
+In order to allow clients to update independently of repositories, the repositories need to continue support for old TUF versions for some period of time after updating. In this TAP, repositories support multiple versions by separating versions with breaking changes into different directories. These directories allow a client to choose the most recent metadata they support while allowing for flexibility in how long a client will take to update to a new spec version. Similarly, allowing the client to maintain multiple versions allows the client to use metadata from multiple directories that are not in communication and so do not coordinate an update to the most recent TUF specification.
```
JustinCappos: So this is when a client is older than a repo.  What about the opposite?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,142 @@
...
+  |- metadata files
+```
+
+## Changes to TUF clients
+
+TUF clients must make a few changes to support changes to the TUF specification.
+
+In order to find compatible updates on a repository, a client must keep track of its version of the TUF specification. To do so, a global variable or other local storage option should contain the client spec version. For simplicity, this field should be formatted according to Semantic Versioning so that it can be directly compared to the spec version in root metadata.
+
+For compatibility, TUF clients may maintain old versions of the client that support previous spec versions. These old versions can be used if the client downloads metadata from a repository or delegated role that does not support the current TUF specification. To allow for this behavior, when a new version of the TUF client is implemented it may contain the ability to call certain functions from the old TUF client for parsing and validating metadata.
```
JustinCappos: > For compatibility, TUF clients may maintain old versions of the client

What does "of the client" refer to here?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,458 @@
...
+the TUF specification, the client should use the version that corresponds to
+each repository to validate its metadata. Results from all repositories, could
+then be compared. The update will only be considered valid if valid metadata
+from both repositories point to the same target file, and it matches the hashes
+provided by each repository. Note that different TUF versions may use different
+hashing algorithms. If this is the case, both hashes should be verified
+independently.
+
+
+### Changes to delegations
```
trishankatdatadog: Delegatees can use new TUF code that implement new TUF specifications, and the repository can simply store their metadata in different directories as suggested above, no?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,458 @@
...
+verified using the major version of the TUF client that corresponds with the
+major version of the root file.
+
+So, if the currently trusted root file is named 4.root.json and uses version
+1.0.0 and the highest supported version is 3.0.0, the client will look for
+5.root.json first in 3.0.0, then 2.0.0, then 1.0.0. If this file is found, the
+client will look for 6.root.json using the same process. To facilitate this, the
+client should maintain functions to parse root files from previous spec
+versions. If the client does not support the spec version of a root file, the
+client shall terminate the update and report the spec version mismatch.
```
trishankatdatadog: One thing missing from this discussion is how to update root metadata across different versions of the TUF specification. If the repo goes from 4.root.json (1.0.0) to 5.root.json (2.0.0) to 6.root.json (3.0.0), shouldn't the repo also produce, say, versions 5 and 6 of root.json in the 1.0.0 directory if, say, 4 and 5 have expired by now?

trishankatdatadog:
Thanks very much for this detailed and well thought-out PR @mnm678! I just have a few comments...

jhdalek55:
Thanks for a well thought out review

mnm678:
Thanks for the review @trishankatdatadog

candidate-tuf-versions.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Changes to TAPs
+TAPs shall be tied to a version of the TUF specification. Once a TAP is accepted the header should be updated to include the first TUF version that will include the TAP. The Preamble Header description in TAP 1 shall be updated to include this field.
+
+## How a repository updates
+
+When a repository manager chooses to update the repository to a new major TUF spec version, they create a new directory on the repository named for the major version (for example 2.0.0). This directory will contain all metadata files with the new spec version. After creating the directory, the repository creates and signs root, snapshot, timestamp, and top level targets metadata using the new TUF spec version and places these metadata files in the directory. Clients will now be able to use these metadata files once their TUF spec versions are also updated.
+
```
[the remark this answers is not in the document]
mnm678: I don't think I understand the question. The metadata files on the repository are still controlled by the same entity, they are just separated into directories by the TUF specification version they support.

candidate-tuf-versions.md
mnm678: I added some description of this to the "How a repository updates" section

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+* Author: Marina Moore, Justin Cappos
+* Status: Draft
+* Content-Type: text/markdown
+* Created: 19-December-2018
+
+# Abstract
+
+The TUF specification does not currently support breaking changes or changes
+that are not backwards compatible. If a repository and a client are not using
+the same version of the TUF specification, metadata can not be safely and
```
JustinCappos:
```suggestion
the same version of the TUF specification, differences in metadata format may mean that metadata cannot be safely and
```

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+reliably verified. This TAP addresses this clash of versions by allowing TUF
+implementations to independently manage updates on clients and repositories.
+This in turn ensures that TUF will continue functioning after breaking changes
+are implemented.
+
+To manage breaking changes without loss of functionality, this TAP requires two
```
JustinCappos: > without loss of functionality

What functionality and for whom?  What is the goal?

You probably need something early on that shows an example of changes that are of each type (breaking, forwards compatible, patch notes).

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+recent specification version supported by both the client and the repository.
+
+
+# Motivation
+
+Various TAPs, including TAPs 3 and 8, propose changes that are not backwards
+compatible. Because these changes are not compatible with previous TUF versions,
+current implementations that use the existing specification can not access the
+new features in these TAPs. By creating a way to deal with non-backwards
+compatible (breaking) changes, TUF will be able to handle a variety of use
```
JustinCappos: What do you mean by handle here?  The goal is still a bit unclear.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+cases, including those that appear below.
+
+## Use case 1: A repository updates to a new TUF spec version
+
+As new features become available in the TUF specification, repositories may wish
+to upgrade to a new version. This could include adding TAPs with breaking
```
JustinCappos: Need to have defined "breaking changes" at this point.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+changes. When the repository upgrades, clients that use an older version of the
+TUF spec will no longer be able to safely parse metadata from the repository.
```
JustinCappos: What does it mean to safely parse?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+time or other constraints.
+
+## Use case 2: A client updates to a new TUF spec version
+
+Just as a repository may be upgraded, a TUF client may wish to upgrade to a new
+TUF spec version to use new features. When the client implements an upgrade that
+includes a breaking change, it cannot be sure that all repositories have also
+upgraded to the new specification version. Therefore, after the client upgrades
+they must ensure that any metadata downloaded from a repository was generated
+using the new version. If the repository is using an older version, the client
```
JustinCappos: Why does the metadata need to be the latest version?  Do newer versions always break old clients?  Can't the client understand both versions?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+specification than the repository hosting the top level roles. Delegated roles
+and the top level repository may be managed by people in different organizations
+who are not able to coordinate upgrading to a new version of the TUF
+specification. In this case, a client should be able to parse delegations that
+generate metadata with a different TUF specification version than the repository.
+
+## Use case 4: A client downloads metadata from multiple repositories
+
+As described in TAP 4, TUF clients may download metadata from multiple
+repositories. These repositories do not coordinate with each other, and so may
```
JustinCappos:
```suggestion
repositories. These repositories may not coordinate with each other, and so may
```

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+not upgrade to a new TUF specification version at the same time. A client should
+be able to use multiple repositories that do not use the same version of TUF.
+For example, a client may download metadata from one repository that uses TUF
+version 1.0.0 and another repository that uses version 2.0.0.
+
+## Use case 5: Allowing Backwards Compatibility
+
+Existing TUF clients will still expect to download and parse metadata without
```
JustinCappos: I'm confused about what you mean here.  I don't understand the case you are trying to convey.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+repositories. Before this TAP there was no method for determining compatibility
+between a client and a repository, so existing TUF repositories should continue
+to distribute metadata using their existing method to support existing clients. This
+means that this mechanism for allowing backwards compatible updates must itself
+be backwards compatible.
+
+# Rationale
+
+We propose this TAP because as TUF continues to evolve, the need for TUF clients
+and repositories to upgrade will grow. This sets up the potential for clients
```
JustinCappos:
```suggestion
and repositories to upgrade their TUF metadata format will grow. This sets up the potential for clients
```

candidate-tuf-versions.md
```diff
@@ -0,0 +1,468 @@
...
+directories, a client is able to choose the most recent metadata they support.
+More details about this directory structure are contained in the [specification](#how-a-repository-updates).
+
+On the client side, this TAP also requires maintenance of multiple versions that
+support different TUF specification
+versions to allow for communication with various repositories. To do so, it is
+recommended that clients maintain functions that can be used to validate
+metadata from previous TUF specification versions. These functions allow a
+client to maintain old versions of the specification while still supporting the
+most recent version.
```
JustinCappos: How does one handle different versions of metadata?  Can you have targets files from different versions than the root, snapshot, etc.?  What files may be different?

mnm678:
@JustinCappos I updated the TAP to make some of the definitions and goals more clear

trishankatdatadog:
Hi, this was opened at the end of 2018. Is there any plan to move it to Draft anytime soon? :)

JustinCappos:
Interesting perspective.  I'd like for those who are interested to have a
chat and see if we can understand your perspective better and perhaps
reconcile the differences.  Our thought when working on this TAP was we
were strictly making it better for implementers and maintainers, but if
there is some potential we could be doing the opposite, we should better
understand how...

candidate-tuf-versions.md
```diff
@@ -0,0 +1,510 @@
...
+This TAP adopts the third option, and describes procedures for both clients and
+repositories to implement multiple TUF specification versions at the same time.
+This requires both clients and
+repositories to maintain multiple versions of the TUF spec, as well as changes
+to the way repositories store metadata and clients parse this metadata.
+
+For repositories, this means continuing support for old TUF versions for some
+period of time after upgrading. This grace period gives existing clients that
+implement old versions of the TUF specification time to implement support for a
+new specification version. Repositories achieve this using a directory structure
```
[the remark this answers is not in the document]
mnm678: My initial thought was that this communication would happen out of band (a blog post/announcement/etc). But looking at it again, it might be nice to add an optional 'deprecated on' field that the repository can use to indicate that a specification version will no longer be supported after a certain date.

mnm678:
> I have mixed feelings about this TAP. On one hand, it seems like a well thought out and logical approach to the problem – kudos for that Marina. On the other, do we really want/need this to be part of the specification? As the specification moves away from recommending wire formats and even repository representations, should it seek to encode concepts of version management? Of course, with my reference implementation maintainer hat on – this would be a complexity and maintenance burden increase in the implementation.
> 
> Perhaps a route forward would be to make an advisory TAP for version management and ensure there's sufficient metadata in the specification, and sufficient discipline around version management in the specification, to support it?

Let's talk about this more in the TAP meeting. Specification version management and backwards compatibility was a bit of a challenge for the reference implementation, so I think it would be nice to at least have some formal guidance about managing spec versions (especially as we move forward with TAPs that include breaking changes). Because a few of these recommendations require changes to TAPs and metadata storage, I think it will be easier to include those changes in the specification. However, if implementations have their own method for version management I don't think these recommendations should supersede those preferences.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,458 @@
...
+For the repositories this means continuing support for old TUF versions for some
+period of time after upgrading. This grace period gives the client time to
+upgrade to a new specification version. Versions with breaking changes are kept
+in separate directories that allow a client to choose the most recent metadata
+they support. To save space, target files should remain in the parent directory
+on the repository. The metadata files (in directories according to their spec
+  version) can point to target files relative to the parent directory.
+
+On the client side, this TAP also requires maintenance of multiple specification
+versions to allow for communication with various repositories. To do so, it is
```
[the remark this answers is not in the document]
mnm678: Yes. The client can choose to only support one version, but for compatibility with multiple repositories with different versions or with delegated targets with a different version than the top-level metadata the client may maintain multiple versions at the same time.

candidate-tuf-versions.md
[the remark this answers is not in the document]
joshuagl: Agreed, I think it makes sense for the repository to be able to indicate the difference between metadata which has expired and metadata which is intentionally no longer updated.

That might be a field for a deprecation timestamp for a given metadata version, or a "supported versions" metadata field that is replicated across metadata versions?

joshuagl:
I've been thinking about this quite a bit whilst offline and now disagree vehemently with myself from yesterday (re-learning a lesson about reviewing). This does need to be in the specification.

I completely agree that this makes things easier for integrators. The reference implementation would want to implement something along these lines, so that we can support our adopters in updating to newer versions of the specification as we improve the standard. I didn't have any doubts about that.

Whilst there is a complexity increase for implementers, it is manageable. In the case of the reference implementation this further necessitates the refactor we have been thinking about and validates some of the early design proposals there.

Thinking about some of our major adopters that have multiple repositories and/or multiple clients interacting with the repositories (i.e. Docker/Notary and PyPI) I realised that if we don't include this in the specification, we make things potentially trickier for implementers who have to be aware of the advisory TAP, and for implementers who might not recognise the need to upgrade specification compatibility throughout the life of their repository.

In short: the complexity increase seems more than worthwhile. Implementers and integrators, particularly those with more controlled update systems where there's only one repository and only one client, can always choose not to handle multiple versions of the specification.

Apologies for the noise, I hope you all don't mind me doing so much of my thinking "out loud" here.

JustinCappos:
> Apologies for the noise, I hope you all don't mind me doing so much of my
> thinking "out loud" here.
>

No, this is really great!  We want to have this sort of rationale /
thinking captured somewhere.  I'm sure that others may have had similar
thoughts and the fact that your thought process is written down clearly may
help others come to the same conclusion.

>

candidate-tuf-versions.md
```diff
@@ -0,0 +1,546 @@
...
+
+A repository may continue to support old major TUF specification versions by creating
+metadata in both the old location and the new directory. A repository manager
+may maintain as many versions as desired, though any version with security
+concerns should be phased out as soon as possible. This can be done by stopping
+the creation of new metadata files in the directory for the phased out version.
+In order to allow clients to parse the root metadata chain, root metadata files
+shall not be deleted even once a version is deprecated.
+
+A repository may indicate the planned phase out of a major version using an
```
mnm678, line 290: @joshuagl I added a deprecation field as discussed. A couple of notes:
* I decided to add the field to targets metadata. I considered root, but changes to root metadata are usually harder to manage (with several offline keys). Also, the targets metadata can be used to indicate deprecation both for top-level metadata and for delegated targets metadata. 
* I'm not a big fan of the name `deprecation_timestamp`, so am open to any clearer or shorter name suggestions

joshuagl: Not putting this in root seems sensible. Is Targets the best role to include this in? Would the Snapshot make more sense? Or are you trying to avoid having this in metadata which will often be signed by online keys?

Regarding the name, would `becomes_obsolete` work better? It's not shorter, but may be clearer?

mnm678: I went with targets instead of snapshot/timestamp mostly to allow the field to differ in delegated targets. The only attack that I can think of on this field would be a denial of service (or convincing a client to upgrade early, but it's hard to imagine that being a successful attack), so the online keys shouldn't be too much of a problem. However in this design delegated targets may use a different TUF specification version than the top-level metadata, so having this field for each targets metadata file seems useful. For example an individual delegated targets file could stop supporting version 1.x before the top-level metadata, so this field would allow them to indicate this change to the client.

I like `becomes_obsolete`, I'll update that in the text.

joshuagl: Ah, yes. I forgot about the desire to support different metadata versions in delegated targets. 

I was thinking it might make more sense in an earlier role in the [client workflow](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#the-client-application). If we have root, timestamp or snapshot metadata which has expired a client can't know whether that's because the metadata is obsolete, or whether something else happened (because we check the expiration on root, timestamp and snapshot and abort if the expiration has passed long before we ever download targets).

My hope was that this metadata field would
1) inform clients when a version of the metadata which hasn't expired is due to be deprecated – this use-case appears to be supported by having `becomes_obsolete` in targets metadata 
2) allow clients to figure out whether metadata that has expired is due to deprecation or something more worrisome – this use-case does not seem to be supported by having `becomes_obsolete` in targets metadata

mnm678: We could also add this field to the timestamp metadata to address case 2 (with the caveat that this would not help if root metadata expires). This would also ensure that the client checks this field even if no target files have been updated (when they might not download targets metadata).

joshuagl: I still have reservations about this not being in the root metadata – how do clients detect unintentionally expired root metadata (or an attack) vs. obsolete root metadata? Or, phrased differently, how does a client know when updates may be available, just at a different end-point, versus updates being unavailable? 

I acknowledge your concerns about including `becomes_obsolete` in the root metadata being harder to manage. Updating metadata specification versions is a big thing, but should it require a quorum of the root signers and their offline keys? 🤷 


The reason I keep belabouring this point is because, to my mind, this is an aspect of TUF's freshness security design principle. The TUF website [states that](https://theupdateframework.io/security/#freshness):

> a client should be able to recognize that updates may exist that they haven't been able to obtain.

I'd argue that's not possible if the client can't detect the difference between root metadata expiring and root metadata being marked as obsolete.

JustinCappos: I've read and thought about this and I can see why we're having an extended discussion on this point.  I think this is a hard decision. 

If you believe the repository is the arbiter of truth and trust, then keeping it in the root role makes sense.  This does effectively let the repository's root role deprecate large swaths of targets files though.  This can be good or bad.

If you think that the developer keys (targets metadata files) are what should be trusted, then having a field in targets makes sense.  (I would argue for version number instead of timestamp since there isn't a global clock / sync requirement for most of TUF except perhaps to print a warning about outdated timestamp metadata.)  This allows parties that delegate make the choice about what versions of metadata are alright.  I do feel like this option means that no one will ever use this and that old metadata will be kept around indefinitely, because I just can't see developers worrying about using this flag...

mnm678: > how do clients detect unintentionally expired root metadata (or an attack) vs. obsolete root metadata? 

Root metadata is designed to expire fairly infrequently, so clients would have a long time (potentially a year or more) to notice the `becomes_obsolete` field in timestamp metadata. However, I agree that updating the specification should be done with the knowledge of the root signers, so maybe including this field in root might be reasonable as it is rarely updated.

> If you believe the repository is the arbiter of truth and trust, then keeping it in the root role makes sense. This does effectively let the repository's root role deprecate large swaths of targets files though. This can be good or bad.

I think regardless of whether this field goes in timestamp or root, it also needs to exist in targets metadata for this reason. In this design delegated targets metadata can use a different specification version than the top-level roles, so they need their own way to indicate deprecation. This is especially important when targets metadata is replicated across multiple repositories.

> I would argue for version number instead of timestamp since there isn't a global clock / sync requirement for most of TUF except perhaps to print a warning about outdated timestamp metadata.

This may be another argument in favor of putting this field in root. Root metadata is always versioned whereas timestamp metadata is only versioned if consistent snapshots are used.

joshuagl: > (I would argue for version number instead of timestamp since there isn't a global clock / sync requirement for most of TUF except perhaps to print a warning about outdated timestamp metadata.)

Is that true? The [detailed client workflow](https://github.com/theupdateframework/specification/blob/a765b3e1275cd023d5fe11ad08fb296c3eb95ff3/tuf-spec.md#5-detailed-workflows) in the latest spec states:

> 1.8. Check for a freeze attack. The latest known time should be lower than the expiration timestamp in the trusted root metadata file (version N). If the trusted root metadata file has expired, abort the update cycle, report the potential freeze attack. On the next update cycle, begin at step 0 and version N of the root metadata file.

I'm not sure how checking for a `becomes_obsolete` date in the metadata is any different from the requirement here to abandon an update if the root metadata file appears to have expired?

mnm678: Thinking about this more, I think that a timestamp might be more useful to clients than a version number as it allows them to make a plan for upgrading, especially in airgapped environments or anywhere else when upgrading might take some planning.

So, per this discussion I will update the TAP to include the becomes_obsolete field in root and targets, and leave its contents as the timestamp for now. This field is optional so if updating root or using a timestamp is an issue, it could be left out.

candidate-tuf-versions.md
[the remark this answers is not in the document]
trishankatdatadog: Sounds a little bit onerous. Surely you don't mean every MINOR or PATCH release every? Only for every MAJOR release, no? So, e.g., `1.0.0` and `2.0.0` are understood to serve every MINOR or PATCH releases under them respectively?

candidate-tuf-versions.md
trishankatdatadog: What do you mean by "clients"? I don't think you mean the TUF reference implementation itself, but clients that use them in turn, right?

candidate-tuf-versions.md
```diff
@@ -0,0 +1,510 @@
...
+compatible metadata.
+
+For clarity, throughout this TAP *upgrading* refers to the process of moving
+from one TUF specification version to another while *updating* refers to the
+process of downloading and validating packages as specified by TUF. The rest of
+this section details the changes to the TUF specification recommended by this
+TAP.
+
+
+## Version Number Format
```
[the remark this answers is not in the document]
trishankatdatadog: Ok, let's link to that and save space if we can.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+open source projects that makes the meaning of a TUF version number change
+consistent and easily understood. The Backwards Compatibility section of a TAP
+should be used to determine whether the TAP creates a breaking change. If the
+change is not backwards compatible, such as those in TAPs 3 and 8, then it will
+be part of a new major version of the TUF specification. Non backwards
+compatible versions add features that change the way that TUF processes updates
+and so clients and repositories need to use code that supports the same major
+version when performing an update in order to maintain security and functionality.
+
+If the change adds a new feature that is backwards compatible, for example in
```
trishankatdatadog, line 224: Could you list an example backwards-compatible new feature that qualifiers for a minor version? I'm hard-pressed to think of one right now.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+top-level metadata to indicate when the metadata they are responsible for will
+no longer support a given TUF specification version. With the addition of this
+field, the "signed" portion of targets metadata will include the following:
+
+```
+{ "_type" : "targets",
+  "spec_version" : SPEC_VERSION,
+  "version" : VERSION,
+  "expires" : EXPIRES,
+  ("becomes_obsolete": BECOMES_OBSOLETE),
```
trishankatdatadog, line 312: What is the `BECOMES_OBSOLETE` schema? Also, seems weird not to group `spec_version` and `becomes_obsolete` together; they don't seem to be useful on their own.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+For existing TUF clients to continue operation after this TAP is implemented,
+repositories may store metadata from before TUF 2.0.0 in the top-level
+repository (with no directory named 1.0.0). This allows existing clients to
+continue downloading metadata from the repository. So a TUF repository that
+upgrades from version 1.0.0 to version 2.0.0 may look like:
+
+
+```
+- Targets files
+- 1.0.0 metadata files
```
trishankatdatadog, line 357:
```suggestion
- 1.0.0
  |- 1.0.0 metadata files
```

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+- 2.0.0
+  |- 2.0.0 metadata files
+```
+
+## Changes to TUF clients
+
+TUF clients must store the TUF specification versions they support and may add
+functions to maintain old versions. In order to find compatible updates on a
+repository, a client must keep track of the TUF specification versions it
+supports. To do so, a global variable or other local storage option should
```
trishankatdatadog, line 367: Let's not recommend global variables, please.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+`becomes_obsolete` field. If this field is present, the client should do
+the following:
+* If the current time is after the time listed in `becomes_obsolete`, the
+client should report to the user that the update is not available due to a
+version mismatch and consider the current metadata file invalid. If
+this is the root or top-level targets file, the update should be terminated.
+* Otherwise, the client should warn the user that the TUF specification version
+they are using will be deprecated at the time indicated by `becomes_obsolete`
+and proceed with the update.
+
```
trishankatdatadog, line 449:
```suggestion
```

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+implementation of TUF and that no security flaws are introduced in code.
+In addition, supply chain security techniques (for example [in-toto](https://in-toto.io/in-toto.html))
+should be used to ensure that vulnerabilities are not introduced into TUF
+clients between when they are written and installed.
+
+Note that an attacker on the same network could inject false metadata to block a
+client from updating. However, this denial of service attack is always possible
+for an attacker on the network with the ability to alter network traffic.
+
+# Backwards Compatibility
```
trishankatdatadog, line 558: I think you should mention how current repositories and clients should continue to keep their current metadata and targets directories, for clients that predate this TAP.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+## Root Key Rotation
+The TUF version upgrade system described in this TAP uses the root key rotation
+mechanism to ensure that the client only uses valid TUF metadata. The first time
+a new TUF version is used by the client, it checks the repository's root
+metadata in the new version's format. If the root metadata is not signed by the
+previous trusted root key, the update does not proceed, and the client halts the
+update. In this way, updates to the repository's TUF version do not impact the
+security of an update.
+
+## How a client upgrades
```
trishankatdatadog, line 538: That is a new level of meta. Love it.

candidate-tuf-versions.md
```diff
@@ -0,0 +1,570 @@
...
+### Delegations
+
+The TUF specification version(s) supported by a delegation does not need to
+match the version(s) supported by the top-level metadata. A delegation may use
+the directory structure described above to indicate which TUF specification
+versions it supports. The TUF client is responsible for parsing the metadata
+from delegations using code that implements the appropriate specification
+version (found using the same procedure as described in
+[Changes to the update process](#changes-to-the-update-process)). If there is no
+compatible specification version between the client and delegation, the client
```
trishankatdatadog, line 500: Just to be clear, you mean matching the MAJOR version, no?

candidate-tuf-versions.md
mnm678: The metadata files actually go in the top level directory to allow for backwards compatibility with existing 1.0.0 implementations that do not know about this directory structure.

candidate-tuf-versions.md
mnm678: Yes, this means the client that is downloading from a TUF repository. I think I define this above, but I can make it clearer here as well.

candidate-tuf-versions.md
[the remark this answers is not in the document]
mnm678: I added a link to the Reference Implementation section of the TAP. Do you think I should remove the description here as well?

candidate-tuf-versions.md
mnm678: TAPs 13 and 4 are both backwards-compatible new features. TAP 13 is an especially good example because it is a change that only affects the client.

mnm678:
Thanks for the review @trishankatdatadog! I responded to a couple of comments inline.

@JustinCappos You're still labeled as requesting a change, is there anything else I should address before this gets merged as a draft?

mnm678:
> This looks ready to assign a TAP number, rename and merge as draft.

Done! I think the number 14 avoids any conflicts.

PR #108, opened 2019-02-11
TAP3: Resolve ambiguity in multi-role delegation

awwad:
- Fix some old text that said that all roles in a multi-role
delegation should agree (instead of a threshold of roles).
- Provide instruction for how to resolve a conflict in a
multi-role delegation with low threshold (e.g. 2/4).


awwad:
Going once, going twice, gone.  @JustinCappos

JustinCappos:
Belated LGTM.  2 hours isn't enough time to review a PR!

PR #111, opened 2019-09-16
tap*: Use TUF-Version header as spec reference

SantiagoTorres:
The TUF-Version header is not utilized as of now. Instead of leaving it
as an unused appendix, utilize it to reference to the version of the
specification that integrates this TAP.

In addition, update all taps that should have a TUF-Version header in
them.

### DIFF
--- a/tap1.md
+++ b/tap1.md
@@ -105,7 +105,7 @@ The format of a TAP is specified with a Content-Type header. The acceptable valu
 
 The Created header records the date that the TAP was assigned a number, while Post-History is used to record the dates of when new versions of the TAP are posted to the TUF mailing list.  Both headers should be in dd-mmm-yyyy format, e.g. 14-Aug-2001.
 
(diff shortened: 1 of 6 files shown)


SantiagoTorres:
I'm opening this as a Draft PR because there's no such thing as an RFC PR :)

tap1.md
```diff
@@ -105,7 +105,7 @@ The format of a TAP is specified with a Content-Type header. The acceptable valu
 
 The Created header records the date that the TAP was assigned a number, while Post-History is used to record the dates of when new versions of the TAP are posted to the TUF mailing list.  Both headers should be in dd-mmm-yyyy format, e.g. 14-Aug-2001.
 
-TAPs will typically have a TUF-Version header which indicates the version of TUF that the feature will be released with. TAPs without a TUF-Version header indicate interoperability standards that will initially be supported through external libraries and tools, and then supplemented by a later TAP to add support to the Reference Implementation.
+TAPs will typically have a TUF-Version header which indicates the version of TUF that the feature will be released with. TAPs without a TUF-Version header indicate interoperability standards that will initially be supported through external libraries and tools, and may eventually become part of the Core Specification. TAPs that refer to processes or recommendations do not require a TUF-Version header either.
```
mnm678: What are interoperability standards? They seem to be for adding features to the reference implementation rather than to the standard, in which case I don't think they belong in TAPs.

PR #112, opened 2020-03-19
Add TAP for keyid flexibility

mnm678:
This TAP proposes a specification change to allow for increased keyid flexibility. The need for this TAP arose from discussion about keyid use cases (https://github.com/theupdateframework/tuf/issues/848, https://github.com/theupdateframework/tuf/issues/660).

A previous draft of this document was available [here](https://docs.google.com/document/d/1ytP9jqaTTTovNjbvk7c6TZGwFzCNUJrpCQGGdAyYwFk/).

### DIFF
--- a/candidate-keyid-tap.md
+++ b/candidate-keyid-tap.md
@@ -11,50 +11,55 @@
 
 # Abstract
 
-The TUF specification requires that every keyid used in TUF metadata
-be calculated using a SHA-256 hash of the public key it represents. This
-algorithm is used elsewhere in the TUF specification and so provides an existing
-method for calculating unique keyids. Yet, such a rigid requirement may be
-challenging to support in a given programming language or system if no existing
-library supports SHA-256 or if the hashes are too long for bandwidth constrained
-applications. For these reasons, this TAP proposes a change to the TUF
-specification that would remove the requirement that all keyids be calculated
-using SHA-256. Instead, the specification will allow metadata owners to use any
-method for calculating keyids as long as each one is unique within the metadata
-file in which it is defined to ensure a fast lookup of trusted signing keys.
-This change will give metadata owners flexibility in how they determine keyids.
+Keyids are used in TUF metadata as shorthand references to identify keys. They
+are used in place of keys in metadata to assign keys to roles and to identify
+them in signature headers. The TUF specification requires that every keyid used
+in TUF metadata be calculated using a SHA2-256 hash of the public key it
+represents. This algorithm is used elsewhere in the TUF specification and so
+provides an existing method for calculating unique keyids. Yet, such a rigid
+requirement does not allow for the deprecation of SHA2-256. A security flaw in
+SHA2-256 may be discovered, so TUF implementers may choose to deprecate this
+algorithm. If SHA2-256 is deprecated in TUF, it should no longer be used to
+calculate keyids. Therefore this TAP proposes a change to the TUF specification
+that would remove the requirement that all keyids be calculated using SHA2-256.
+Instead, the specification will allow metadata owners to use any method for
+calculating keyids as long as each one is unique within the metadata file in
+which it is defined to ensure a fast lookup of trusted signing keys. This
+change will allow for the deprecation of SHA2-256 and will give metadata owners
+flexibility in how they determine keyids.
+
 
 # Motivation
 
-Currently, the TUF specification requires that keyids must be the SHA-256 hash
+Currently, the TUF specification requires that keyids must be the SHA2-256 hash
 of the public key they represent. This algorithm ensures that keyids are unique
 within a metadata file (and indeed, throughout the implementation) and creates a
-short, space-saving representation. SHA-256 also offers a number of secure
+short, space-saving representation. SHA2-256 also offers a number of secure
 hashing properties, though these are not necessary for these purposes. In this
-case SHA-256 is simply a way to calculate a unique identifier employing an
+case SHA2-256 is simply a way to calculate a unique identifier employing an
 algorithm that is already in use by the system.
 
 The specification sets the following requirements for keyid calculation:
-1. The KEYID of a key is the hexdigest of the SHA-256 hash of the canonical JSON form of the key.
+1. The KEYID of a key is the hexdigest of the SHA2-256 hash of the canonical JSON form of the key.
 2. Clients MUST calculate each KEYID to verify this is correct for the associated key.
-3. Clients MUST ensure that for any KEYID … only one unique key has that KEYID.
+3. Clients MUST ensure that for any KEYID only one unique key has that KEYID.
 
 ## Problems with this requirement
-Mandating that keyids be calculated using SHA-256 has created a number of issues
+Mandating that keyids be calculated using SHA2-256 has created a number of issues
 for some implementations, such as:
 * Lack of consistency in implementations that use other hash algorithms for
-  calculating file hashes and would prefer not to introduce SHA-256 for this one
+  calculating file hashes and would prefer not to introduce SHA2-256 for this one
   instance. For example, the PEP 458 implementation (https://python.zulipchat.com/#narrow/stream/223926-pep458-implementation)
-  will use the * Blake 2 hashing algorithm throughout the implementation.
+  will use the BLAKE2 hashing algorithm throughout the implementation.
 * Incompatibility with some smart cards and PGP implementations that have their
   own way of calculating keyids.
-* Inability to adapt if SHA-256 should be deprecated. In such a case, metadata
+* Inability to adapt if SHA2-256 should be deprecated. In such a case, metadata
   owners may decide that maintaining a deprecated algorithm for use in keyid
   calculation does not make sense.
-* Space concerns may require even shorter hashes than those SHA-256 can generate,
+* Space concerns may require even shorter hashes than those SHA2-256 can generate,
   such as an index.
 In these and other cases, TUF should provide a metadata file owner with the
-flexibility to use keyids that are not calculated using SHA-256.
+flexibility to use keyids that are not calculated using SHA2-256.
 
 # Rationale
 
@@ -95,7 +100,7 @@ to maliciously sign files.
 # Specification
 
 With just a few minor changes to the current TUF specification process, we can
-remove the requirement that keyids must be calculated using SHA-256. First, the
+remove the requirement that keyids must be calculated using SHA2-256. First, the
 specification wording should be updated to allow the metadata owner to calculate
 keyids using any method that produces a unique identifier within the metadata
 file. This means replacing requirements 1 and 2 above with a description of
@@ -137,11 +142,11 @@ calculation method.
 ## Keyid Deprecation
 With the proposed specification changes, the method used to determine keyids
 is not only more flexible, but it may be deprecated using the following process
-for each key K and keyid k in the root or delegating targets metadata file:
-* The owner of the metadata file determines a new keyid h for K using the new method.
-* In the next version of the metadata file, the metadata owner replaces k with h
+for each key K and keyid I in the root or delegating targets metadata file:
+* The owner of the metadata file determines a new keyid J for K using the new method.
+* In the next version of the metadata file, the metadata owner replaces I with J
   in the keyid definition for K.
-* Any files previously signed by K should list h as the keyid instead of k.
+* Any files previously signed by K should list J as the keyid instead of I.
   These files do not need to be resigned as only the signature header will be updated.
 Once this process is complete, the metadata owner is using a new method to
 determine the keyids used by that metadata file.
@@ -156,17 +161,17 @@ be unique for each delegated role. It is possible for different keyids to
 represent the same key in different metadata files, even if both metadata files
 delegate to the same role. Consider two delegated targets metadata files A and B
 that delegate to the same targets metadata file C. If A delegates to C with
-key k with keyid g and B delegates to C with key k with keyid h, both of these
+key k with keyid x and B delegates to C with key k with keyid y, both of these
 delegations can be processed during the preorder depth-first search of targets
 metadata as follows:
-* When the search reaches A, it will look for a signature with a keyid of g in C.
+* When the search reaches A, it will look for a signature with a keyid of x in C.
   If it finds this and validates it, the search will continue if a threshold of
   signatures has not been reached.
-* When the search reaches B, it will look for a signature with a keyid of h in C.
+* When the search reaches B, it will look for a signature with a keyid of y in C.
   If it finds this and validates it, the search will continue if a threshold of
   signatures has not been reached.
 Once the search is complete, if a threshold of signatures is reached the
-metadata in C will be used to continue the update process. Therefore, g and h
+metadata in C will be used to continue the update process. Therefore, x and y
 may be used as keyids for k in different metadata files. So that clients can
 validate signatures using each of these keyids, they both must be used to
 identify a valid signature using k in C’s header. As clients store keyids only
@@ -186,13 +191,13 @@ collisions and protects clients from privilege escalation attacks.
 
 # Backwards Compatibility
 
-Metadata files that are generated using SHA-256 will be compatible with clients
+Metadata files that are generated using SHA2-256 will be compatible with clients
 that implement this change. However, clients that continue to check that
-keyids are generated using SHA-256 will not be compatible with metadata that
+keyids are generated using SHA2-256 will not be compatible with metadata that
 uses a different method for calculating keyids.
 
 For backwards compatibility, metadata owners may choose to continue to use
-SHA-256 to calculate keyids.
+SHA2-256 to calculate keyids.
 
 # Augmented Reference Implementation
 
--- a/candidate-keyid-tap.md
+++ b/candidate-keyid-tap.md
@@ -19,7 +19,7 @@ represents. This algorithm is used elsewhere in the TUF specification and so
 provides an existing method for calculating unique keyids. Yet, such a rigid
 requirement does not allow for the deprecation of SHA2-256. A security flaw in
 SHA2-256 may be discovered, so TUF implementers may choose to deprecate this
-algorithm. If SHA2-256 is deprecated in TUF, it should no longer be used to
+algorithm. If SHA2-256 is deprecated in a TUF implementation, it should no longer be used to
 calculate keyids. Therefore TUF should allow more flexibility in how keyids are
 determined. To this end, this TAP proposes a change to the TUF specification
 that would remove the requirement that all keyids be calculated using SHA2-256.
@@ -126,7 +126,7 @@ In order for TUF clients to adhere to these specification changes, they may have
 to change the way they store and process keyids. Clients will use the keyids
 from a metadata file only for all delegations defined in that metadata file. So
 if a targets metadata file T delegates to A and B, the client should verify the
-signatures of A and B using the trusted keyids in from T. When verifying
+signatures of A and B using the trusted keyids from T. When verifying
 signatures, clients should try all signatures that match their trusted keyid(s).
 If T trusts keyid K to sign A’s metadata, the client should check all
 signatures in A that list a keyid of K. This means that if another metadata file
--- a/candidate-keyid-tap.md
+++ b/candidate-keyid-tap.md
@@ -200,6 +200,13 @@ may be used as keyids for D in different metadata files. As clients store keyids
 only for use in the current delegation, this should not require a change to the
 client process described in this document.
 
+In this case, the same key D is used to verify C twice, once when A delegates to
+C and once when B delegates to C. As the key is only used once during each
+delegation, this does not violate the client verification of key uniqueness
+described in this TAP. If the keyids L and K were both used in the same
+delegation (say A delegating to C), then these signatures would only contribute
+a single valid signature to the threshold due to the client verification.
+
 It is also possible for the same keyid to represent different keys in different
 metadata files. Consider a targets metadata file A that delegates to C with key
 D and keyid K and a targets metadata file B that delegates to C with key E and
--- a/candidate-keyid-tap.md
+++ b/candidate-keyid-tap.md
@@ -118,7 +118,10 @@ keys to verify signatures.” To ensure that a client does not use the same key
 to verify a signature more than once, they must additionally check that all keys
 applied to a signature threshold are unique. So, the specification should
 additionally require that "Clients MUST use each key only once during a given
-signature verification." All metadata definitions would remain the same, but
+signature verification." During this de-duplication check, the client should use
+a standardized format for keys, like the [modulus and exponent for RSA](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#numbers)
+or the [point and curve for ECC](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicNumbers).
+All metadata definitions would remain the same, but
 the client’s verification process would track keyids within each metadata file
 instead of globally.
 
@@ -133,7 +136,8 @@ signatures in A that list a keyid of K. This means that if another metadata file
 M delegates to A, it would be able to use the same keyid with a different key.
 However, clients must ensure that duplicate keys are not applied to the same
 signature threshold. To do so, they must additionally keep track of the keys
-used to verify a signature. Once the signatures for A and B have been checked,
+used to verify a signature using a standardized format as discussed above. Once
+the signatures for A and B have been checked,
 the client no longer needs to store the keyid mapping listed in T. During the
 preorder depth-first search of targets metadata, the keyids from each targets
 metadata file should be used in only that stage of the depth-first search.
--- a/candidate-keyid-tap.md
+++ b/candidate-keyid-tap.md
@@ -41,9 +41,9 @@ case SHA2-256 is simply a way to calculate a unique identifier employing an
 algorithm that is already in use by the system.
 
 The specification sets the following requirements for keyid calculation:
-1. The KEYID of a key is the hexdigest of the SHA2-256 hash of the canonical JSON form of the key.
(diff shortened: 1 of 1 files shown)


candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+within a metadata file (and indeed, throughout the implementation) and creates a
+short, space-saving representation. SHA-256 also offers a number of secure
+hashing properties, though these are not necessary for these purposes. In this
+case SHA-256 is simply a way to calculate a unique identifier employing an
+algorithm that is already in use by the system.
+
+The specification sets the following requirements for keyid calculation:
+1. The KEYID of a key is the hexdigest of the SHA-256 hash of the canonical JSON form of the key.
+2. Clients MUST calculate each KEYID to verify this is correct for the associated key.
+3. Clients MUST ensure that for any KEYID … only one unique key has that KEYID.
```
trishankatdatadog:
```suggestion
3. Clients MUST ensure that for any KEYID, only one unique key has that KEYID.
```

candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Problems with this requirement
+Mandating that keyids be calculated using SHA-256 has created a number of issues
+for some implementations, such as:
+* Lack of consistency in implementations that use other hash algorithms for
+  calculating file hashes and would prefer not to introduce SHA-256 for this one
+  instance. For example, the PEP 458 implementation (https://python.zulipchat.com/#narrow/stream/223926-pep458-implementation)
+  will use the * Blake 2 hashing algorithm throughout the implementation.
```
trishankatdatadog:
```suggestion
  will use the * BLAKE2 hashing algorithm throughout the implementation.
```

candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+* Status: Draft
+* Content-Type: markdown
+* Created: 18-03-2020
+* TUF-Version: 1.1.0
+* Post-History: <dates of postings to the TUF mailing list>
+
+# Abstract
+
+The TUF specification requires that every keyid used in TUF metadata
+be calculated using a SHA-256 hash of the public key it represents. This
```
trishankatdatadog:
```suggestion
be calculated using a SHA2-256 hash of the public key it represents. This
```

trishankatdatadog: Just to be super clear, let's replace all instances of `SHA-256` with `SHA2-256`, please.

candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+that implement this change. However, clients that continue to check that
+keyids are generated using SHA-256 will not be compatible with metadata that
+uses a different method for calculating keyids.
+
+For backwards compatibility, metadata owners may choose to continue to use
+SHA-256 to calculate keyids.
+
+# Augmented Reference Implementation
+
+TODO
```
trishankatdatadog: This needs to be finished, ofc 🙂

candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+scheme independent of the client implementation, root and targets metadata may
+use different methods to determine keyids, especially if they are managed by
+different people (ie TAP 5). In addition, the repository may update the scheme
+at any time to deprecate a hash algorithm or change to a different keyid
+calculation method.
+
+## Keyid Deprecation
+With the proposed specification changes, the method used to determine keyids
+is not only more flexible, but it may be deprecated using the following process
+for each key K and keyid k in the root or delegating targets metadata file:
```
trishankatdatadog: Maybe try not to use both `K` and `k`, please? Makes it hard to read...

candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+* The owner of the metadata file determines a new keyid h for K using the new method.
+* In the next version of the metadata file, the metadata owner replaces k with h
+  in the keyid definition for K.
+* Any files previously signed by K should list h as the keyid instead of k.
```
trishankatdatadog: Also, maybe not `h`

candidate-keyid-tap.md
```diff
@@ -0,0 +1,203 @@
...
+metadata file remain unique. Metadata owners should only publish metadata that
+contains a unique keyid to key mapping.
+
+## Implications for complex delegation trees
+Although keyids need to be unique within each metadata file, they do not need to
+be unique for each delegated role. It is possible for different keyids to
+represent the same key in different metadata files, even if both metadata files
+delegate to the same role. Consider two delegated targets metadata files A and B
+that delegate to the same targets metadata file C. If A delegates to C with
+key k with keyid g and B delegates to C with key k with keyid h, both of these
```
trishankatdatadog: Maybe choose easier to read keyids? Making them lowercase makes them hard to distinguish

mnm678: Are the new variable names easier to distinguish? All the letters looked surprisingly similar when I was trying to find a suitable replacement.

trishankatdatadog: It's slightly better, but there's still some lowercase here and uppercase there. Could we use a consistent example throughout?

trishankatdatadog:
I don't have write access to a few TUF repos. Can someone help me fix this? @lukpueh @JustinCappos

candidate-keyid-tap.md
```diff
@@ -0,0 +1,208 @@
...
+* Any files previously signed by D should list L as the keyid instead of K.
+  These files do not need to be resigned as only the signature header will be updated.
+Once this process is complete, the metadata owner is using a new method to
+determine the keyids used by that metadata file.
+
+As keyid deprecation is executed, it is important that keyids within each
+metadata file remain unique. Metadata owners should only publish metadata that
+contains a unique keyid to key mapping.
+
+## Implications for complex delegation trees
```
erickt: You touch upon this in the specification, but it might be helpful to also include an example here for the case where the targets metadata file A delegates to C using the key D and the keyid K, as well as another targets metadata B delegates to C using the key E and the keyid K. From what I understand, this should result in C being signed with:

```
{
    "sigs": [
        {
             "keyid": "K",
             "sig": "1234..."
        },
        {
             "keyid": "K",
             "sig": "abcd..."
        },
        ...
    ],
    ...
}
```
Both signatures will be checked if the delegation search ever ends up at A or B.

candidate-keyid-tap.md
```diff
@@ -0,0 +1,208 @@
...
+# Specification
+
+With just a few minor changes to the current TUF specification process, we can
+remove the requirement that keyids must be calculated using SHA2-256. First, the
+specification wording should be updated to allow the metadata owner to calculate
+keyids using any method that produces a unique identifier within the metadata
+file. This means replacing requirements 1 and 2 above with a description of
+required keyid properties, ie “The KEYID is an identifier for the key that is
+determined by the metadata owner and MUST be unique within the root or
+delegating targets metadata file.” Once this keyid is determined by the metadata
```
erickt: > The KEYID is an identifier for the key that is determined by the metadata owner and MUST be unique within the root or delegating targets metadata file.

Now that there wouldn't be a cryptographically derived keyid for each key, it's now possible to list the same key multiple times, each with a different keyid. Can we instead impose another rule that the key must also be unique? Without this rule, an attacker could duplicate a signature to reduce the effective threshold of keys needed to validate the metadata. We do this in go-tuf and rust-tuf to avoid double counting keys due to our (temporary) support of python-tuf's `keyid_hash_algorithms` field.

JustinCappos: This is a good point.  If there are multiple delegations to the same KEYID as part of a threshold, is this allowed?  If I delegate to both Alice and Bob, who both delegate to Charlie, is Charlie's approval enough?  (I think I know the answer in both cases and this is mostly tangental to this issue, but think we should discuss if we think TUF is doing the right thing.)

candidate-keyid-tap.md
```diff
@@ -0,0 +1,208 @@
...
+If T trusts keyid K to sign A’s metadata, the client should check all
+signatures in A that list a keyid of K. This means that if another metadata file
+M delegates to A, it would be able to use the same keyid with a different key.
+Once the signatures for A and B have been checked, the client no longer needs to
+store the keyid mapping listed in T. During the preorder depth-first search of
+targets metadata, the keyids from each targets metadata file should be used in
+only that stage of the depth-first search.
+
+These changes to the specification would allow the repository to use any scheme
+to determine keyids without needing to communicate it to clients. By making this
```
JustinCappos: I don't understand how or why the repository is determining KEYIDs.  Can you clarify?

candidate-keyid-tap.md
```diff
@@ -0,0 +1,208 @@
...
+scheme independent of the client implementation, root and targets metadata may
+use different methods to determine keyids, especially if they are managed by
+different people (ie TAP 5). In addition, the repository may update the scheme
+at any time to deprecate a hash algorithm or change to a different keyid
+calculation method.
+
+## Keyid Deprecation
```
JustinCappos: I don't really understand why this is needed anymore.  Just generate a new file with whatever new KEYID scheme you want.  All consumers didn't understand how you got the KEYIDs anyways, right?

candidate-keyid-tap.md
```diff
@@ -0,0 +1,208 @@
...
+Currently, the TUF specification requires that keyids must be the SHA2-256 hash
+of the public key they represent. This algorithm ensures that keyids are unique
+within a metadata file (and indeed, throughout the implementation) and creates a
+short, space-saving representation. SHA2-256 also offers a number of secure
+hashing properties, though these are not necessary for these purposes. In this
+case SHA2-256 is simply a way to calculate a unique identifier employing an
+algorithm that is already in use by the system.
+
+The specification sets the following requirements for keyid calculation:
+1. The KEYID of a key is the hexdigest of the SHA2-256 hash of the canonical JSON form of the key.
```
JustinCappos: Not the biggest thing, but we likely should be consistent with keyid vs KEYID capitalization. If you're quoting, that's fine, perhaps be more explicit about that in your formatting though...

candidate-keyid-tap.md
mnm678: The process is pretty straightforward, but the metadata owner should ensure that the keyids remain unique throughout any transition. This might not need a whole section, but I think that it's important to note.

candidate-keyid-tap.md
mnm678: Added, thanks for the suggestion!

candidate-keyid-tap.md
mnm678: I added a description of this in the scenarios below. As currently written, in this situation Charlie's approval would be enough. I think that this should be sufficient to trust the metadata as a threshold of signatures using unique keys must be reached to obtain Charlie's approval.

candidate-keyid-tap.md
```diff
@@ -0,0 +1,275 @@
...
+
+Keyids are used in TUF metadata as shorthand references to identify keys. They
+are used in place of keys in metadata to assign keys to roles and to identify
+them in signature headers. The TUF specification requires that every keyid used
+in TUF metadata be calculated using a SHA2-256 hash of the public key it
+represents. This algorithm is used elsewhere in the TUF specification and so
+provides an existing method for calculating unique keyids. Yet, such a rigid
+requirement does not allow for the deprecation of SHA2-256. A security flaw in
+SHA2-256 may be discovered, so TUF implementers may choose to deprecate this
+algorithm. If SHA2-256 is deprecated in TUF, it should no longer be used to
```
joshuagl, line 22: It can be confusing differentiating between the spec and an implementation, I think you mean an implementation here?
```suggestion
algorithm. If SHA2-256 is deprecated in a TUF implementation, it should no longer be used to
```

candidate-keyid-tap.md
joshuagl: I have less practical experience with tuf than the other folks in this discussion, but I would be surprised that Charlie's approval would be enough. My understanding of the specification is that thresholds are a defence against key compromise. In the scenario described we only have to compromise a single key in order to meet a threshold that is > 1.

candidate-keyid-tap.md
```diff
@@ -0,0 +1,275 @@
...
+additionally require that "Clients MUST use each key only once during a given
+signature verification." All metadata definitions would remain the same, but
+the client’s verification process would track keyids within each metadata file
+instead of globally.
+
+In order for TUF clients to adhere to these specification changes, they may have
+to change the way they store and process keyids. Clients will use the keyids
+from a metadata file only for all delegations defined in that metadata file. So
+if a targets metadata file T delegates to A and B, the client should verify the
+signatures of A and B using the trusted keyids in from T. When verifying
```
joshuagl, line 129: Remove extra in
```suggestion
signatures of A and B using the trusted keyids from T. When verifying
```

candidate-keyid-tap.md
```diff
@@ -0,0 +1,275 @@
...
+file and in all signatures that use the corresponding key. When parsing metadata
+signatures, the client would use the keyid(s) listed in the signature header to
+find the key(s) that are trusted for that role in the delegating metadata. This
+should be described in the specification by replacing requirement 3 above with
+“Clients MUST use the keyids from the delegating role to look up trusted signing
+keys to verify signatures.” To ensure that a client does not use the same key
+to verify a signature more than once, they must additionally check that all keys
+applied to a signature threshold are unique. So, the specification should
+additionally require that "Clients MUST use each key only once during a given
+signature verification." All metadata definitions would remain the same, but
```
lukpueh, line 121: So @mnm678 and I [discussed this quite a bit](https://docs.google.com/document/d/1ytP9jqaTTTovNjbvk7c6TZGwFzCNUJrpCQGGdAyYwFk/edit?disco=AAAAJRGh1fw), and I seem to remember that we agreed that a client-side key or keyid uniqueness check within a delegation only prevents scenarios, where the repository owner already screwed up, i.e. uses the same keyid for different keys, or different keyids for one key.

Or am I missing something? 

If not, I wonder if we really want to make client verification more complex, because the client can't trust repositories to do their job right.  I mean, the client also doesn't check for faux-pas like the delegation scenario, Justin described above, or for a repository using the same key to sign root and timestamp.

candidate-keyid-tap.md
JustinCappos: > or he at least does not allow them to further delegate trust (IIUC, this is what the "terminating" flag in delegations is for).

terminating means something different.  It means, if using a terminating delegation to Alice, then if Alice (and the people she delegates to) doesn't know the answer, terminate the search and say it is unknown.

Alice and Bob may have reasons for thinking that delegating to Charlie is fine.  If we added an option where Alice and Bob's delegations have to eventually need to result in different keys signing a file, then you'd need to resolve *all* keys that both Alice and Bob trust.  This would be a bit slower and more complex than what we're currently doing. 

I'm not saying that this is impossible or even necessarily that we shouldn't do it, but it does make delegation resolution more complex.

candidate-keyid-tap.md
mnm678: There are a couple of different uniqueness requirements that the keyid/key systems needs. I agree that where possible we want to simplify the client workflow by trusting information found in signed metadata files. The situations in which uniqueness is needed include:
- keyids provide a mapping that can be used to determine which key is trusted. As @lukpueh and I discussed in the above link, this mapping is provided in signed metadata, and so can be trusted by the client without additional verification.
- Duplicate keys cannot be applied to the same threshold. If an attacker is able to append a duplicate signature, the client should only apply this signature to the threshold once. In this case, I think that the client can do a O(1) check (eg by adding each key to a set) to ensure uniqueness. Previously this was done by ensuring the keyid was only used once, but using the key instead seems like a less error-prone way to make this check (that solves some [current](https://github.com/theupdateframework/tuf/issues/372) [issues](https://github.com/theupdateframework/tuf/issues/920)).
- In a multi-role delegation each role's approval is only used once. This is the situation discussed in the comments above. I think that like the keyid/key mapping, this situation can only come up if the delegation tree allows two roles to delegate to the same role. Afaik there is not a way for an attacker without a threshold of keys to create this situation, so it should be left up to the repository manager to occasionally audit to ensure the delegation tree makes sense.

candidate-keyid-tap.md
lukpueh: > terminating means something different. It means, if using a terminating delegation to Alice, then if Alice (and the people she delegates to) doesn't know the answer, terminate the search and say it is unknown.

Oh right. Thanks for clarifying, @JustinCappos! `terminating=True` terminates the search horizontally, not vertically, is that a way to put it? 

> Alice and Bob may have reasons for thinking that delegating to Charlie is fine.

Agreed. Furthermore, Alice and Bob don't necessarily know about each other, nor that Justin requires a threshold of 2, so they might subvert it by accident. On the other hand, how likely is it that they both delegate to the same role unknowingly?

> If we added an option where Alice and Bob's delegations have to eventually need to result in different keys signing a file, then you'd need to resolve _all_ keys that both Alice and Bob trust. This would be a bit slower and more complex than what we're currently doing.

I'm not sure about that. Doesn't the DFS traversal in this case look something like this:
```
requirement: Justin requires two roles in agreement about a given target
branch search a): Justin asks Alice, Alice asks Charlie, Charlie has target, back up to Justin
branch search b): Justin asks Bob, Bob asks Charlie, Charlie has target, back up to Justin
success: Justin has two roles in agreement and releases the target
```

Shouldn't it be enough to, along with Charlies knowledge about the target, also pass up the public key and let Justin only release the target if two roles with different keys were in agreement?

By the way, and to make sure we are on the same page, when we say threshold in this context, we mean `min_roles_in_agreement` as described in TAP 3, right? (see [related conversation](https://github.com/theupdateframework/specification/pull/57/files#r340734215) about disambiguation)

Also note that multi-role delegation as specified in TAP 3 is neither part of the reference implementation nor part of the current specification (https://github.com/theupdateframework/specification/pull/57). And I also think the issue we are discussing here exists independently of Marina's keyid proposal.

candidate-keyid-tap.md
lukpueh: Thanks for the extra clarification, @mnm678! This makes sense to me. Regarding the third item, I left some thoughts in the discussion above.

candidate-keyid-tap.md
JustinCappos: > ```
> requirement: Justin requires two roles in agreement about a given target
> branch search a): Justin asks Alice, Alice asks Charlie, Charlie has target, back up to Justin
> branch search b): Justin asks Bob, Bob asks Charlie, Charlie has target, back up to Justin
> success: Justin has two roles in agreement and releases the target
> ```
> 
> Shouldn't it be enough to, along with Charlies knowledge about the target, also pass up the public key and let Justin only release the target if two roles with different keys were in agreement?

The current algorithm returns with the first match.  So suppose that Alice (or Bob) also delegated trust to Daniela, who also approved the target.  You should approve it in this case, but the current algorithm would not do so.  You'd need to match *all* parties that agree.  Then you'd need to be sure that Alice and Bob each have at least one person that was not yet chosen who they could select to fulfill this need.  Also, what if Alice and Bob have threshold delegations to Charlie and Daniela in this case?  It's just a lot messier to deal with, in my view.

candidate-keyid-tap.md
trishankatdatadog: How would we keep track of duplicate public keys? I mean, what if there are the same public keys, but listed in different formats? (I know we use only PEM right now, but humor me.)

mnm678: I would guess that if multiple formats are supported, they would need to be converted to a consistent format for signature verification, so the uniqueness check would take place after any conversion.

Are there any use cases in which multiple formats would be used but not be able to be converted? In this case we might need to have some more discussion.

trishankatdatadog: Yeah. This was already an issue before, but it's something we're forced to think through now thanks to this TAP, which is not as simple as I thought it would be...

lukpueh: Hah! The exact same concern occurred to me yesterday evening.  Get out of my after-work hours head, @trishankatdatadog!! :P

As @mnm678 points out,  de-duplicating on a serialized format such as PEM is a bad idea. Attackers could take advantage of optional PEM headers, whitespace, etc. 

I think, we should de-duplicate on the basic components of public keys, e.g. [modulus and exponent for RSA](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#numbers); [modulus, sub-group order, generator and public value for DSA](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/dsa/#numbers) and  [point and curve for ECC](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicNumbers).

mnm678: Thanks for bringing this up @trishankatdatadog and @lukpueh. I added a couple of sentences about ensuring a standard format for keys. I think the exact format can be left to the implementer as long as it uniquely identifies the key.

candidate-keyid-tap.md
```diff
@@ -118,7 +118,10 @@ keys to verify signatures.” To ensure that a client does not use the same key
 to verify a signature more than once, they must additionally check that all keys
 applied to a signature threshold are unique. So, the specification should
 additionally require that "Clients MUST use each key only once during a given
-signature verification." All metadata definitions would remain the same, but
+signature verification." During this de-duplication check, the client should use
+a standardized format for keys, like the [modulus and exponent for RSA](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#numbers)
```
trishankatdatadog, line 122:
```suggestion
a standardized representation for keys, like the [modulus and exponent for RSA](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#numbers)
```

candidate-keyid-tap.md
```diff
@@ -133,7 +136,8 @@ signatures in A that list a keyid of K. This means that if another metadata file
 M delegates to A, it would be able to use the same keyid with a different key.
 However, clients must ensure that duplicate keys are not applied to the same
 signature threshold. To do so, they must additionally keep track of the keys
-used to verify a signature. Once the signatures for A and B have been checked,
+used to verify a signature using a standardized format as discussed above. Once
```
trishankatdatadog, line 139:
```suggestion
used to verify a signature using a standardized representation as discussed above. Once
```

candidate-keyid-tap.md
trishankatdatadog: Thanks, Marina! As I noted in my comments, I think we should say "representation" instead of "format", because that's what we mean. Even if the same RSA key has two different public key formats, it has the same modulus and exponent.

candidate-keyid-tap.md
mnm678: I think that this issue pertains more to the existing delegation resolution algorithm than to the keyid calculation, so I propose we open a new issue to discuss this issue further. As @lukpueh mentioned, TAP 3 is not currently included in the specification, so this issue does not exist with the current specification.

candidate-keyid-tap.md
lukpueh: But once again (and thanks for pointing this out, @SantiagoTorres), if the attacker does not control the metadata that lists the public keys, then they cannot make different representations of the same key each count towards the threshold.
And if the attacker does control the metadata, they can add any key they want and uniqueness checks on the client are moot. 

This means that we really only need to de-duplicate on the one key representation in the delegating metadata. Unless, and this brings me back to my initial comment above, we don't trust the delegating role.

Sorry for adding to the confusion with my comment above, I was briefly confused too. Maybe we should re-hash the threat model for this document.

trishankatdatadog: Yes, I was more interested in someone (e.g., the delegating role) trying to deliberately fool `python-tuf` into accepting what are actually duplicate keys by using different public key formats which actually describe the same public key parameters.

SantiagoTorres: Right, but this delegating role could also just change the threshold? My impression is that this is necessary because we want to avoid footguns, but it's not a very exploitable attack vector.

trishankatdatadog: Sure, but it's still something to think about, I think, especially if the client-side checks are not expensive.

mnm678: The unique key check already exists to ensure that an attacker cannot append duplicate signatures (although this requirement is not explicitly stated in the specification). The main difference here is requiring that the client do this uniqueness check on the key instead of the keyid. If all goes correctly, using the key does not make any difference, but if the metadata accidentally has duplicate keys with different keyids, using keyids to verify uniqueness of keys could lead to a key being applied to a threshold more than once.

candidate-keyid-tap.md
mnm678: On a closer look at this issue, it appears that the ``visited`` flag ensures that Charlie's role will not be visited more than once in the DFS as described in 4.4.1 of the [client workflow](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows). Does this solve the issue @JustinCappos @lukpueh?

JustinCappos: I don't believe this solves it.  I don't think it would handle the case above with Daniela, for example.

I do think it would be fine to move this elsewhere, but I think the most important question is what should the system actually do?  I think we could code it to work however, but I think we should definitely strive to do what a user would expect first and foremost.

mnm678: Ok, I created a [new issue](https://github.com/theupdateframework/specification/issues/96) to continue this discussion.

mnm678:
@JustinCappos @trishankatdatadog Are there any blockers to getting this merged as a draft?

trishankatdatadog:
> @JustinCappos @trishankatdatadog Are there any blockers to getting this merged as a draft?

As a draft? No, I don't think so. Justin, what about you?

JustinCappos:
> @JustinCappos @trishankatdatadog Are there any blockers to getting this merged as a draft?

Not at all.  Feel free.

mnm678:
> > @JustinCappos @trishankatdatadog Are there any blockers to getting this merged as a draft?
> 
> Not at all. Feel free.

It looks like I don't have write access to this repository.

PR #120, opened 2020-07-02
Add candidate TAP for succinct hash bin delegations

mnm678:


candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Specification
+
+This TAP adds the following extension to delegations:
+```
+(“succinct_hash_delegations” : {
+    “prefix_bit_length” : BIT_LENGTH
+})
+```
+Where 2^BIT_LENGTH is the number of bins. BIT_LENGTH must be an integer between 1 and 32 (inclusive).
```
trishankatdatadog, line 80: Why 32? Why not 64? Also, I feel like if we need to accommodate 4,294,967,296 bins, something is wrong.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+
+```
+{ "keys" : {
+       KEYID : KEY,
+       ... },
+   "roles" : [{
+       "name": DELEGATING_ROLENAME,
+       "keyids" : [ KEYID, ... ] ,
+       "threshold" : THRESHOLD,
+       ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
```
trishankatdatadog, line 99: What should an implementation do if a delegation lists both `path_hash_prefixes` and `succinct_hash_delegations`?

candidate-succinct-delegation-tap.md
JustinCappos: > Why 32? Why not 64? Also, I feel like if we need to accommodate 4,294,967,296 bins, something is wrong.

Right, this is exactly the reason.  2**64 is just absurdly large.  We may use a setting like 2**20 or so for DockerHub, so thought 24 was a little too tight.  We wanted to make it divisible by 8 so that it is the first so many bytes at most.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
+* TAP:
```
joshuagl, line 1:
```suggestion
* TAP: 14
```

We have a TAP 13 in draft, let's call this 14?

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+* Last-Modified: 02-07-2020
+* Author: Marina Moore, Justin Cappos
+* Status: Draft
+* Created: 23-06-2020
+* TUF-Version:
+* Post-History:
+
+# Abstract
+
+TUF delegating metadata contains the keyid and path for each delegation, so when a single metadata file delegates to many others the delegating metadata’s size increases proportional to the number of delegations. Clients must download this large delegating metadata file when they update any image it delegates to, therefore large delegating metadata increases the client’s metadata overhead. To reduce the metadata overhead for targets metadata that contains many delegations, TUF supports hashed bin delegations. Hashed bin delegations use targets files that delegate to ‘bin’ roles that delegate to the actual targets. Targets are distributed to bins using the hash of the target filename. Using hashed bin delegations, the delegating metadata only delegates to the bins, and so contain less data. Thus, hashed bin delegations reduce the client’s metadata overhead.
```
joshuagl, line 13:
```suggestion
TUF delegating metadata contains the keyid and path for each delegation, so when a single metadata file delegates to many others the delegating metadata’s size increases proportional to the number of delegations. Clients must download this large delegating metadata file when they update any target it delegates to, therefore large delegating metadata increases the client’s metadata overhead. To reduce the metadata overhead for targets metadata that contains many delegations, TUF supports hashed bin delegations. Hashed bin delegations use targets files that delegate to ‘bin’ roles that delegate to the actual targets. Targets are distributed to bins using the hash of the target filename. Using hashed bin delegations, the delegating metadata only delegates to the bins, and so contain less data. Thus, hashed bin delegations reduce the client’s metadata overhead.
```

Changes image->target in "when they update any image it delegates to"

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+       "name": alice.hbd-03,
+       "keyids" : [ abc123 ] ,
+       "threshold" : 1,
+       "path_hash_prefixes" : [ 03*],
+        "paths" : [ “path/to/directory/” ],
+       "terminating": false,
+   },... ]
+ }
+ ```
+Every client will then have to download this large delegating metadata file. Note that all of the information in italics above is exactly the same in each delegation. The only data that differs between these delegations are the bolded fields, the name and path_hash_prefix. The name has a common prefix (underlined), so only differs by a count and the path_hash_prefix is generated using the hash algorithm and number of bins.
```
joshuagl, line 62: The formatting (italics, bold, underline) has all been lost and GitHub flavoured Markdown does not support underlines.

I'd suggest we don't format the repeated information, italicise the differing fields and bold the common prefix.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Rationale
+
+Hashed bin delegations commonly use the same key for each bin in order to allow for automated signing of targets files. If the same key is used for all bins, this key should only need to be listed once in the delegating metadata.
+
+Similarly, bins are usually named using a numbering system with a common prefix (i.e. alice.hbd-0, alice.hbd-1, …). The delegating metadata only needs to describe this naming scheme and the location that these metadata files are stored.
+
+The delegating metadata could significantly reduce the client’s metadata overhead by providing a succinct description of the keyid and prefix instead of repeating these for each delegation.
```
joshuagl, line 70: I think the impact of this TAP would be emphasised by some numbers here. I think we could generate some fairly easily?

candidate-succinct-delegation-tap.md
trishankatdatadog: I understand why 64 is too large. My point is that 32 is too large in practice too. I suppose nothing should, in principle, stop an implementation from doing this, but I just can't see a practical reason for it.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+
+The names of each bin will be determined by the bin number and the name of the delegating entity. It will be structured as DELEGATING_ROLENAME.hbd-COUNT where DELEGATING_ROLENAME is the name of the role that delegated to the hash bins and COUNT is a value 0-2^BIT_LENGTH-1 that represents the bin number.
+
+If a delegation contains a succinct hash delegation, all files represented by this delegation must exist on the repository, even if they do not contain any targets or delegations. These bin files should be uploaded before the metadata that delegates to them.
```
joshuagl, line 88:
```suggestion
If a delegation contains a succinct hash delegation, all metadata files represented by this delegation must exist on the repository, even if they do not contain any targets or delegations. These bin files should be uploaded before the metadata that delegates to them.
```

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+ }
+ ```
+
+# Security Analysis
+
+This TAP will not negatively affect the security of TUF. The succinct_hash_delegations field provides a shorter way to describe an existing feature when it follows a predictable pattern. If the hashed bin delegations do not follow this pattern (they use different keys for different bins, etc), they may instead use the existing hash bin mechanism.
+
+Repositories that use access control for file uploading should take hashed bin delegations into consideration. Upload access for the name DELEGATING_ROLENAME.hbd-* should have the same permissions as DELEGATING_ROLENAME.
+
+If a repository has multiple delegations to an image, clients will resolve these using prioritized delegations. So if a delegation uses succinct hash bins and another role delegates to DELEGATING_ROLENAME.hbd-\*, the client will use the delegation with a higher priority.
```
joshuagl, line 115:
```suggestion
If a repository has multiple delegations to a target, clients will resolve these using prioritized delegations. So if a delegation uses succinct hash bins and another role delegates to DELEGATING_ROLENAME.hbd-\*, the client will use the delegation with a higher priority.
```

image->target

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Backwards Compatibility
+
+This TAP is not backwards compatible, and will need to be included in a major version release of the specification. We consider the access control that must be supported on the repository, as well as compatibility between the client and the repository.
+
+The repository must ensure that the correct access control mechanisms are applied to filenames of the form DELEGATING_ROLENAME.hbd-\*. As discussed in the security analysis, this metadata file should only be uploaded by DELEGATING_ROLENAME. The repository should handle this access control before succinct hash bin delegations are used so that other uploaders are not able to use the DELEGATING_ROLENAME.hbd-* filename for targets.
+
+In order for succinct hash bin delegations to be used, both the delegation and the client must understand the succint_hash_delegations field. Consider a client Bob who wants to download an image J and an uploader Alice who is responsible for delegating to J. This is what would happen if one or both of them supports succinct_hash_delegations.
```
joshuagl, line 123:
```suggestion
In order for succinct hash bin delegations to be used, both the delegation and the client must understand the succint_hash_delegations field. Consider a client Bob who wants to download a target J and an uploader Alice who is responsible for delegating to J. This is what would happen if one or both of them supports succinct_hash_delegations.
```

image -> target

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+
+| Parties that support succinct_hash_delegations | Result |
+| --- | --- |
+| Neither Alice nor Bob support succinct_hash_delegations | Alice would not use succinct_hash_delegations to delegate to J, and so Bob would be able to download and verify the image using the existing mechanisms. |
```
joshuagl, line 127:
```suggestion
| Neither Alice nor Bob support succinct_hash_delegations | Alice would not use succinct_hash_delegations to delegate to J, and so Bob would be able to download and verify the target using the existing mechanisms. |
```

image -> target

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+| Alice supports succint_hash_delegations, Bob does not | Alice may use succinct_hash_delegations to delegate to any of her images, including J. Bob will download metadata that has the succinct_hash_delegations field and will not be able to find the delegation that points to J. |
```
joshuagl, line 128:
```suggestion
| Alice supports succint_hash_delegations, Bob does not | Alice may use succinct_hash_delegations to delegate to any of her targets, including J. Bob will download metadata that has the succinct_hash_delegations field and will not be able to find the delegation that points to J. |
```

images->targets

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,132 @@
...
+| Bob supports succinct_hash_delegation, Alice does not | Alice will not use succinct_hash_delegations to delegate to J. Bob will not see this field in the targets metadata, and so will look for the delegation using the existing mechanisms |
+| Both Alice and Bob support succinct_hash_delegations | Alice may use succinct_hash_delegations to delegate to her images, including J. Bob will see the succinct_hash_delegations field in targets metadata and will download the alice.hdb-x bin metadata file that corresponds to J. |
```
joshuagl, line 130:
```suggestion
| Both Alice and Bob support succinct_hash_delegations | Alice may use succinct_hash_delegations to delegate to her targets, including J. Bob will see the succinct_hash_delegations field in targets metadata and will download the alice.hdb-x bin metadata file that corresponds to J. |
```

images->targets

candidate-succinct-delegation-tap.md
mnm678: In practice, implementations will probably not need 2^32 bins, but I don't see why we can't allow them. If repositories grow in size over time (especially very large repositories like DockerHub) they shouldn't have to worry about the threshold that we set here. Having this threshold very large means we won't need another breaking change to raise this threshold later.

candidate-succinct-delegation-tap.md
mnm678: Great question! I will add text to the tap. In theory it should be possible to include both `path_hash_prefixes` and `succinct_hash_delegations`, but the keyids, etc would be shared across these delegations. I think it will be more straightforward to say that the `succinct_hash_delegations` have priority and so if this field is present, `path_hash_prefixes` should be ignored.

candidate-succinct-delegation-tap.md
mnm678: Good catch, it looks like this didn't translate well. I just pushed a change that uses markdown syntax highlighting in the code block, but the * and _'s still show up. Does anyone know of a better way to format that in GitHub Markdown?

joshuagl: Ugh, that's unfortunate. The Markdown code block is interpreted as wanting to show the Markdown markup, then the formatting is also applied on top.

It looks like the only way to achieve the formatting we want is to use HTML tags directly...

```
<pre><code>
    “delegations”:{ "keys" : {
           abc123 : abcdef123456,
           },
       "roles" : [{
           "name": <b><u>alice.hbd-</u>00</b>,
           "keyids" : [ abc123 ] ,
           "threshold" : 1,
           "path_hash_prefixes" : [ 00* ],
            "paths" : [ “path/to/directory/” ],
           "terminating": false,
       },
```
becomes

<pre><code>
    “delegations”:{ "keys" : {
           abc123 : abcdef123456,
           },
       "roles" : [{
           "name": <b><u>alice.hbd-</u>00</b>,
           "keyids" : [ abc123 ] ,
           "threshold" : 1,
           "path_hash_prefixes" : [ 00* ],
            "paths" : [ “path/to/directory/” ],
           "terminating": false,
       },

mnm678:
> Should this _require_ the unnumbered TAP for [TUF version management](https://github.com/theupdateframework/taps/pull/107/files)? (aside: I plan to review that over the coming days)

This TAP is not backwards compatible, so it would be better if the version management TAP was accepted first to prevent any compatibility problems. However, this TAP doesn't depend on the changes in the other TAP, and would work with a different version management strategy as well.

candidate-succinct-delegation-tap.md
mnm678: I added some numbers based on some [calculations done for Notary v2](https://docs.google.com/spreadsheets/d/18iwWnWvAAZ4In33EWJBgdAWVFE720B_z0eQlB4FpjNc/edit?usp=sharing). There is a lot of other math in that document, so I can make one specific to this TAP if anyone thinks that would help clarify the advantage of this delegation method.

mnm678:
> Would you please briefly describe how these numbers were calculated for the unfamiliar reader? Thanks!

Yes, the current calculations are in a [document with a lot of other information](https://docs.google.com/spreadsheets/d/18iwWnWvAAZ4In33EWJBgdAWVFE720B_z0eQlB4FpjNc/edit?usp=sharing), so I will copy those somewhere else and include a link.

mnm678:
@trishankatdatadog @joshuagl Thanks for the reviews! I believe I addressed all of the comments. Is there anything else before I add a TAP number and merge this as a draft?

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,240 @@
...
+
+The names of each bin will be determined by the bin number and the
+name of the delegating entity. It will be structured as
+DELEGATING_ROLENAME.hbd-COUNT where DELEGATING_ROLENAME is the name of
+the role that delegated to the hashed bins and COUNT is a value 0
+2^BIT_LENGTH-1 that represents the bin number.
+
+The `succinct_hash_delegations` will be prioritized over
+`path_hash_prefixes`. If both of these fields appear in a delegation,
+the `path_hash_prefixes` should be ignored in favor of the
```
trishankatdatadog, line 163:
```suggestion
the `path_hash_prefixes` MUST (as per RFC 2119) be ignored in favor of the
```

trishankatdatadog:
> @trishankatdatadog @joshuagl Thanks for the reviews! I believe I addressed all of the comments. Is there anything else before I add a TAP number and merge this as a draft?

I think we're good. I added one more comment ("MUST" instead of "should", much more prescriptive).

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,240 @@
...
+* Status: Draft
+* Created: 23-06-2020
+* TUF-Version:
+* Post-History:
+
+# Abstract
+
+TUF delegating metadata contains the keyid and path for each
+delegation, so when a single metadata file delegates to many others
+the delegating metadata’s size increases proportional to the number of delegations. Clients must download this large delegating metadata file
```
joshuagl, line 15: NIT: long line here. Also proportional -> proportionally 

```suggestion
the delegating metadata’s size increases proportionally to the number of
delegations. Clients must download this large delegating metadata file
```

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,240 @@
...
+when they update any target it delegates to, therefore large
+delegating metadata increases the client’s metadata overhead. To
+reduce the metadata overhead for targets metadata that contains many
+delegations, TUF supports hashed bin delegations. Hashed bin
+delegations use targets files that delegate to ‘bin’ roles that
+delegate to the actual targets. Targets are distributed to bins using
```
joshuagl, line 21:
```suggestion
delegations, TUF supports hashed bin delegations. Hashed bin delegations
use a targets role that delegates to ‘bin’ roles, which sign metadata
for the actual targets in the bin. Targets are distributed to bins using
```

I'm not sure if this actually makes things clearer?

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,240 @@
...
+
+Similarly, bins are usually named using a numbering system with a
+common prefix (i.e. alice.hbd-0, alice.hbd-1, …). The delegating
+metadata only needs to describe this naming scheme and the location
+that these metadata files are stored.
+
+The delegating metadata could significantly reduce the client’s
+metadata overhead by providing a succinct description of the keyid and
+prefix instead of repeating these for each delegation. For a repository
+with 50,000,000 images using the existing hashed bin delegation
```
joshuagl, line 118:
```suggestion
with 50,000,000 targets using the existing hashed bin delegation
```

candidate-succinct-delegation-tap.md
joshuagl: I see that, per the TAP Workflow in [TAP 1](https://github.com/theupdateframework/taps/blob/master/tap1.md#submitting-a-tap), a TAP should not have a number assigned until it has been accepted as a Draft.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,241 @@
...
+
+As in the current use of hashed bin delegations, target files will be
+distributed to bins based on the SHA2-256 hash of the target path and
+the path_hash_prefixes associated with each bin. The
+path_hash_prefixes will be computed by the repository and client to
+determine which bin each target belongs to. This field will include
+the first BIT_LENGTH bits of the hash and will be determined for each
+bin i by computing i * BIT_LENGTH. Once computed, the
+path_hash_prefixes must be represented as the first BIT_LENGTH bits of
+the hash represented in base16, followed by a “*”. So for a BIT_LENGTH
```
lukpueh, line 151: I think you need to escape asterisks here and elsewhere in the free text, or you mark it up as verbatim text.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,241 @@
...
+of 3, the path_hash_prefix of the first bin would be “0*” (binary
+values starting with 000) , the second bin would be “2*” (binary
+values 001), the third “4*”, then “6*”, “8*”, “a*”, “c*”, and “e*”.
```
lukpueh, line 154: I don't fully understand the bin partitioning maths here. Are you sure that this matches what we [currently do](https://github.com/theupdateframework/tuf/blob/00e15c4714f33dedd9cb5cc5602e12a94261c254/tuf/repository_lib.py#L1060-L1101)? Would you mind elaborating how `BIT_LENGTH` relates to what we currently call `prefix_length`?

lukpueh, line 154: Okay, I just re-read the paragraph and I understand it now. :) The math seems to match what's in the reference implementation, but the binary value in the parentheses for bin "2*" should be 010, right?

Also, I was confused by the word *starting*, do you mean the binary representation of the hash *starts* with 000, or the range of hash path prefixes for bin 0\* *starts* with 000? Either way, I think it would be helpful to list all path prefixes for every bin in this example, it's only 8 bins and 2 prefixes per bins.

And I have another question, why (and where) do we need to represent the `hash_path_prefixes` *as the first BIT_LENGTH bits of the hash represented in base16, followed by a “\*”*? Isn't the point about `succinct_hash_delegations` that we no longer need the `path_hash_prefixes`?

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,241 @@
...
+
+The names of each bin will be determined by the bin number and the
+name of the delegating entity. It will be structured as
+DELEGATING_ROLENAME.hbd-COUNT where DELEGATING_ROLENAME is the name of
+the role that delegated to the hashed bins and COUNT is a value 0
+2^BIT_LENGTH-1 that represents the bin number.
+
+The `succinct_hash_delegations` will be prioritized over
+`path_hash_prefixes`. If both of these fields appear in a delegation,
+the `path_hash_prefixes` MUST (as per RFC 2119) be ignored in favor of the
```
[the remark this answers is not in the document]
mnm678: I think RFC2119 is assumed throughout the specification, so I'll remove that here

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,241 @@
...
+values 001), the third “4*”, then “6*”, “8*”, “a*”, “c*”, and “e*”.
+
+The names of each bin will be determined by the bin number and the
+name of the delegating entity. It will be structured as
+DELEGATING_ROLENAME.hbd-COUNT where DELEGATING_ROLENAME is the name of
+the role that delegated to the hashed bins and COUNT is a value 0
+2^BIT_LENGTH-1 that represents the bin number.
+
+The `succinct_hash_delegations` will be prioritized over
+`path_hash_prefixes`. If both of these fields appear in a delegation,
```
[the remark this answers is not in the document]
mnm678: This allows for some backwards compatibility as a repository is switching from path_hash_prefixes to succinct_hash_delegations, but it might be simpler to mandate that only one or the other can be used.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,241 @@
...
+the `path_hash_prefixes` MUST (as per RFC 2119) be ignored in favor of the
+`succinct_hash_delegations`. If a user wishes to use both of these
+fields, they may do so in separate delegations.
+
+If a delegation contains a succinct hash delegation, all metadata
+files represented by this delegation must exist on the repository,
```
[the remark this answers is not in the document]
mnm678: It was important then too :)

This prevents an attacker from hiding certain bins from a client.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,241 @@
...
+succinct_hash_delegations field provides a shorter way to describe an
+existing feature when it follows a predictable pattern. If the hashed
+bin delegations do not follow this pattern (they use different keys
+for different bins, etc), they may instead use the existing hashed bin
+mechanism.
+
+Repositories that use access control for file uploading should take
+hashed bin delegations into consideration. Upload access for the name
+DELEGATING_ROLENAME.hbd-* should have the same permissions as
+DELEGATING_ROLENAME.
```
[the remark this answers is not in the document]
mnm678: There currently isn't much discussion about access control in the specification, but I thought this was worth mentioning in this case.

candidate-succinct-delegation-tap.md
mnm678: I updated this, but I think the relationship between the binary prefixes and their base64 representations might be a confusing way to explain this (especially because, as you mention, we no longer need to use path_hash_prefixes). I'll revisit this in a bit to see if I can clarify it better.

mnm678: I updated this again and removed all the mentions of path_hash_prefixes.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,233 @@
...
+calculated, see [this spreadsheet](https://docs.google.com/spreadsheets/d/10AKDsHsM2mmh45CWCNFxihJ9f-SP6gXYv7WcWpt-fDQ/edit#gid=0).
+
+# Specification
+
+This TAP adds the following extension to delegations:
+
+```
+("succinct_hash_delegations" : {
+    "prefix_bit_length" : BIT_LENGTH
+})
```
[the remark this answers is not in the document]
lukpueh: `generate_succinct_delegations` sounds a bit like a boolean to me. What about `"delegation_hash_prefix_len": BIT_LENGTH`? We could also keep the dictionary and add an additional optional `role_name_prefix` to allow the delegating role to override the default `DELEGATING_ROLENAME.hbd-` prefix for delegated bin roles. What do you think?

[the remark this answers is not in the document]
joshuagl: I really like the idea of being able to specify tole_name_prefix! Rather than have it be a field in the `succinct_hash_delegation` dictionary, should it be part of the `DELEGATIONS` object so that it can also be used for `path_hash_prefixes` and give us consistently named hashed bin delegations, regardless of which mechanism they were created with?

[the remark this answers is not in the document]
mnm678: I agree that specifying the `role_name_prefix` would be helpful to give the delegating metadata more flexibility. With the existing mechanism, it is already possible to use any role name for hashed bins (although the api does not provide this flexibility). Maybe as part of the implementation we can add the ability to define `role_name_prefix` for both succinct hashed bin delegations (by adding it to the metadata) and for legacy hashed bin delegations (by adding it to the api).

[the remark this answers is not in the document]
lukpueh: Exactly, a `role_name_prefix` in the delegations object is not needed for normal/legacy hashed bin delegations, because each delegated hash bin role name is listed explicitly in the delegation object. But it would definitely be great to support a prefix parameter in the API for any type of hashed bin delegation (see https://github.com/theupdateframework/tuf/pull/1106#issuecomment-679043312)

What do you think about using the `name` field to specify the prefix? It is already re-purposed, by assigning the delegating role name in the case of succinct hbd, and if we specify a prefix we don't need that anymore, right? OTOH, it might be better to aim for semantic correctness and not call it name if it is only a name prefix.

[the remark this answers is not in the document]
mnm678: For simplicity, I like the idea of re-purposing the `name` field. This field doesn't have any other use for succinct hashed bin delegations (as the name is generated from the prefix), so it won't interfere with any other functionality.

If there are no objections, I will update this pr and theupdateframework/tuf#1106 to use `delegation_hash_prefix_len` and describe the use of custom role name prefixes.

candidate-succinct-delegation-tap.md
```diff
@@ -152,9 +151,10 @@ values starting with 000, the second bin would include binary values starting
 with 001, the third 010, then 011, 100, 101, 110, 111.
 
 The names of each bin will be determined by the bin number and the
-name of the delegating entity. It will be structured as
-DELEGATING_ROLENAME.hbd-COUNT where DELEGATING_ROLENAME is the name of
-the role that delegated to the hashed bins and COUNT is a value between 0 and
+PREFIX listed in the `name` field of the delegation. By default, the PREFIX
```
joshuagl, line 154: by default, when the name field is empty? or something else? If the former, does that make the name field optional?

candidate-succinct-delegation-tap.md
```diff
@@ -174,13 +174,11 @@ With the addition of succinct hashed bins, the delegation will contain:
...
    "roles" : [{
-       "name": DELEGATING_ROLENAME,
+       "name": PREFIX,
        "keyids" : [ KEYID, ... ] ,
        "threshold" : THRESHOLD,
        ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
-       ("succinct_hash_delegations" : {
-    "prefix_bit_length" : BIT_LENGTH
-        })
+       ("delegation_hash_prefix_len" : BIT_LENGTH)
```
joshuagl, line 181: I'm mildly concerned that `path_hash_prefixes` and `delegation_hash_prefix_len` are not obviously related. We can address this somewhat by improving wording in the spec, but ideally the names would be more indicative of their relation.

Maybe this is something we should file an issue about to address in a major version update once TAP 14 is Final?

mnm678:
@lukpueh @joshuagl Per our discussion today, I updated the TAP to include a `name_prefix` field to more explicitly describe the generated bins.

candidate-succinct-delegation-tap.md
```diff
@@ -174,11 +176,14 @@ With the addition of succinct hashed bins, the delegation will contain:
        KEYID : KEY,
        ... },
    "roles" : [{
-       "name": PREFIX,
+       "name": ROLENAME,
```
joshuagl, line 179: I thought we agreed to make this optional, because for a succinct hashed bin delegation the field is invalid?

mnm678: We did, I updated the document to make this field optional and add a quick description.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,238 @@
...
+even if they do not contain any target files or delegations. These bin
+files should be uploaded before the metadata that delegates to them.
+With the addition of succinct hashed bins, the delegation will contain:
+
+```
+{ "keys" : {
+       KEYID : KEY,
+       ... },
+   "roles" : [{
+       ("name": ROLENAME,)
```
joshuagl, line 181: I think this effectively becomes
```suggestion
       ("name": ROLENAME |
       "succinct_hash_delegations" : {
         "delegation_hash_prefix_len" : BIT_LENGTH,
         "bin_name_prefix" : NAME_PREFIX}),
```

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,238 @@
...
+       "keyids" : [ KEYID, ... ] ,
+       "threshold" : THRESHOLD,
+       ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+       ("succinct_hash_delegations" : {
+         "delegation_hash_prefix_len" : BIT_LENGTH,
+         "bin_name_prefix" : NAME_PREFIX
+       })
```
joshuagl, line 188: Then this would become:

```suggestion
       ("path_hash_prefixes" : [ HEX_DIGEST, ... ],)
```

joshuagl (commented):
I think the format of the delegations object could be cleaned up, but I'm not sure my proposals here make it any clearer.

If I'm following correctly a delegations object needs either `name` or `succinct_hash_delegations` and `path_hash_prefixes` is optional. If the delegations object includes a `succinct_hash_delegations` field then any `path_hash_prefixes` field is ignored. 

Is that correct? Can we rewrite the object format for the specification to make this clearer?

mnm678:
> If I'm following correctly a delegations object needs either `name` or `succinct_hash_delegations` and `path_hash_prefixes` is optional. If the delegations object includes a `succinct_hash_delegations` field then any `path_hash_prefixes` field is ignored.
> 
> Is that correct? Can we rewrite the object format for the specification to make this clearer?

Yes, and I agree that we should make this more clear in the object format. However, it'd be nice if we could still relate that `succinct_hash_delegations` supersedes `path_hash_prefixes`. I'll think about this some more and see if there's a clear way to relate all of that.

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,238 @@
...
+bin i by taking the first BIT_LENGTH bits of i. So for a BIT_LENGTH
+of 3, the first bin would include binary
+values starting with 000, the second bin would include binary values starting
+with 001, the third 010, then 011, 100, 101, 110, 111.
+
+The rolename of each bin will be determined by the bin number and the
+NAME_PREFIX listed in the `bin_name_prefix` field of the delegation.
+The name will be structured as
+NAME_PREFIX-COUNT where COUNT is a value between 0 and
+2^BIT_LENGTH-1 (inclusive) that represents the bin number. If the
```
jku, line 160: COUNT is hexadecimal or?

jku (commented):
I've not looked at the implementation yet but this looks very useful :+1:

candidate-succinct-delegation-tap.md
```diff
@@ -0,0 +1,238 @@
...
+
+The delegating metadata could significantly reduce the client’s
+metadata overhead by providing a succinct description of the keyid and
+prefix instead of repeating these for each delegation. For a repository
+with 50,000,000 target files using the existing hashed bin delegation
+technique, the snapshot and targets metadata overhead would be around
+1,600,000 bytes for each target. Using succinct hashed bin delegations,
+the snapshot and targets metadata overhead for a target can be reduced
+to about 550,000 bytes. For more detail about how these overheads were
+calculated, see [this spreadsheet](https://docs.google.com/spreadsheets/d/10AKDsHsM2mmh45CWCNFxihJ9f-SP6gXYv7WcWpt-fDQ/edit#gid=0).
```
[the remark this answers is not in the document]
mnm678: I'm open to other ways to present this. I thought it was too many calculations for a markdown table.

I also added some comments to the spreadsheet to clarify those terms.

joshuagl:
> > If I'm following correctly a delegations object needs either `name` or `succinct_hash_delegations` and `path_hash_prefixes` is optional. If the delegations object includes a `succinct_hash_delegations` field then any `path_hash_prefixes` field is ignored.
> > Is that correct? Can we rewrite the object format for the specification to make this clearer?
> 
> Yes, and I agree that we should make this more clear in the object format. However, it'd be nice if we could still relate that `succinct_hash_delegations` supersedes `path_hash_prefixes`. I'll think about this some more and see if there's a clear way to relate all of that.

Looking a this with fresh eyes, I think it's more important in the object definition to tie `succinct_hash_delegations` and `path_hash_prefixes` together. I've been unable to come up with a better alternative. I think the object format you originally proposed is the most sensible.

candidate-succinct-delegation-tap.md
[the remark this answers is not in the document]
joshuagl: It has been a while, are we still planning to repurpose the `name` field as the prefix and drop the `bin_name_prefix`?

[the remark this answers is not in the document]
mnm678: If I remember correctly, we decided against that to make the field names more clear, but I'm still open to using the name field.

mnm678:
@joshuagl @JustinCappos @trishankatdatadog @lukpueh Can we merge this TAP as a draft? I think the only open question is the use of `name` vs `bin_name_prefix`, and that will probably be resolved through some experimenting on the PoC.

trishankatdatadog:
> @joshuagl @JustinCappos @trishankatdatadog @lukpueh Can we merge this TAP as a draft? I think the only open question is the use of `name` vs `bin_name_prefix`, and that will probably be resolved through some experimenting on the PoC.

Yes, let's get this out ASAP for PyPA, maybe a nice Christmas gift!

joshuagl:
Agreed, would be great to get this TAP implemented in the reference implementation for PyPI to use.
I just filed https://github.com/theupdateframework/taps/issues/132 to track outstanding items for this TAP.

PR #123, opened 2020-07-20
TAP 5: mark as rejected

joshuagl:
I've tried to capture the outcome of discussion related to TAP 5 from the [TAP Editors meeting](https://hackmd.io/MlPcGPs-RamfAA6C-Zq1Jw) on Friday in the commit message here.

* Have I correctly captured the issues we discussed?
* Is there a better place to document the reason for rejection than the commit message adding the "Rejected" status?

CC: meeting attendees @mnm678 @trishankatdatadog @JustinCappos @lukpueh

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,13 @@
 
 ## Draft
 
-* [TAP 5: Setting URLs for roles in the root metadata file](tap5.md)
 * [TAP 8: Key rotation and explicit self-revocation](tap8.md)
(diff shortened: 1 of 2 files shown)


JustinCappos:
Looks great!  Would you kindly also include the README change in this?  

(If anyone objects, please say so now before I merge...)

joshuagl:
> Looks great! Would you kindly also include the README change in this?

Thanks. I've updated the patch (force push) to also include a README.md edit.

PR #125, opened 2020-09-17
Add TAP introducing snapshot Merkle trees

mnm678:
This TAP introduces snapshot Merkle trees as a scalable solution to snapshot metadata for large repositories. This grew out of a [discussion of metadata overheads](https://docs.google.com/spreadsheets/d/18iwWnWvAAZ4In33EWJBgdAWVFE720B_z0eQlB4FpjNc/edit?ts=5ed7d6f4#gid=0) associated with the Notary v2 effort.

There is also a draft of the reference implementation available at theupdateframework/tuf#1113.


mnm678:
cc @JustinCappos @joshuagl

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+## Merkle tree generation
+
+When the repository generates snapshot metadata, instead of putting the version
+information for all targets into a single file, it instead uses the version
+information to generate a Merkle tree.  Each target’s version information forms
+a leaf of the tree, then these leaves are used to build a Merkle tree. The
+internal nodes of a Merkle tree contain the hash of the leaf nodes. The exact
+algorithm for generating this Merkle tree (ie the order of leaves in the hash,
+how version information is encoded), is left to the implementer, but this
+algorithm should be documented in a POUF so that implementations can be
```
trishankatdatadog, line 97:
```suggestion
algorithm should be documented in a [POUF](https://github.com/theupdateframework/taps/blob/master/tap11.md) so that implementations can be
```

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Garbage collection
+When a threshold of timestamp keys are revoked and replaced, the repository no
+longer needs to store snapshot Merkle files signed by the previous timestamp
+key. Replacing the timestamp key is an opportunity for fast forward attack
+recovery, and so all version information from before the replacement is no
+longer valid. At this point, the repository may garbage collect all snapshot
+Merkle metadata files.
+
+# Security Analysis
```
trishankatdatadog, line 185: I like the descriptions and analysis of these attacks, although I think it could be made clearer to non-experts by illustrating a short and concrete example. WDYT?

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+they increase it to a sufficiently large number (say the maximum integer value),
+the client will not accept any future version of the target as the version
+number will be below the previous version. To recover from this attack,
+auditors and clients should not check version information from before a
+timestamp key replacement. This allows a timestamp key replacement to be used
+as a reset after a fast forward attack. The existing system handles fast
+forward attack recovery in a similar manner, by instructing clients to delete
+stored version information after a timestamp key replacement.
+
+## Mix and match attack
```
trishankatdatadog, line 217: Yes, but I think it's important to clarify that although mix-and-match attacks by MitM attackers are not possible, they are possible to do by attackers who control timestamp and snapshot keys.

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+* Type: Standardization
+* Status: Draft
+* Content-Type: markdown
+* Created: 14/09/2020
+* +TUF-Version:
+* +Post-History:
+
+ # Abstract
+
+ To optimize the snapshot metadata file size for large registries, registries
```
jku, line 15: registry? is this the same as repository?

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+* Last-Modified: 17/09/2020
+* Author: Marina Moore, Justin Cappos
+* Type: Standardization
+* Status: Draft
+* Content-Type: markdown
+* Created: 14/09/2020
+* +TUF-Version:
+* +Post-History:
+
+ # Abstract
```
jku, line 13: This chapter feels quite heavy on technical detail and light on the high level impact of this TAP

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+
+# Motivation
+
+For very large repositories, the snapshot metadata file could get very large.
+This snapshot metadata file must be downloaded on every update cycle, and so
+could significantly impact the metadata overhead. For example, if a repository
+has 50,000,000 targets, the snapshot metadata will be about 380,000,000 bytes
+(https://docs.google.com/spreadsheets/d/18iwWnWvAAZ4In33EWJBgdAWVFE720B_z0eQlB4FpjNc/edit?ts=5ed7d6f4#gid=0).
+For this reason, it is necessary to create a more scalable solution for snapshot
+metadata that does not significantly impact the security properties of TUF.
```
jku, line 45: The reasoning seems sound but the wording is misleading I think: the issue isn't the number of targets alone but the design decision of 1 target per metadata file? So it would be more accurate to say 

> For example, if a repository
> ￼has 10,000,000 target metadata files, the snapshot metadata will be about 380,000,000 bytes

or did I misunderstand? (it's entirely possible: I'm not able to understand most of the spreadsheet with the context I have)

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,244 @@
...
+invalid Merkle tree that contains a previous version of a target. This attack
+is prevented by both the client’s verification and by auditors. When the client
+verifies previous versions of the snapshot Merkle metadata for a target, they
+ensure that the version number of that target has not decreased. However, if
+the attacker controls the timestamp key(s) and the repository, the previous
+snapshot Merkle metadata downloaded by the client may also be invalid. To
+protect against this case, third party auditors store the previous version of
+all metadata, and will detect when the version number decreases in a new
+Merkle tree. As long as the client checks for an auditor’s verification, the
+client will not install the rolled-back version of the target.
```
jku, line 202: I think this should be clear about how this proposal compares to current situation -- both when clients accept only 3rd party audited trees and when auditors are not used. 

This actually applies to all security impact sections: Clear comparison to status quo would be great

mnm678:
Thanks for the reviews @trishankatdatadog and @jku! I pushed some new commits to address your comments, especially in the security analysis section.

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,276 @@
...
+In addition the path should contain direction information so that the client
+will know whether each node is a left or right sibling when reconstructing the
+tree.
+
+This information will be included in the following metadata format:
+```
+{ “leaf_contents”: {METAFILES},
+  “Merkle_path”: {INDEX:HASH}
+  “path_directions”:{INDEX:DIR}
+}
```
erickt, line 120: I think is worthwhile to scope out how the hash algorithm is specified, in case we wanted to switch implementations over time, both for how a hash tree is computed (for example, I just found out about [Verkle Trees](https://math.mit.edu/research/highschool/primes/materials/2019/conf/12-5-Kuszmaul.pdf)), and the underlying hash algorithm itself.

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,276 @@
...
+the snapshot Merkle metadata to check version information. If this field is
+present, the client will download the snapshot Merkle metadata file only for
+the targets metadata the client is attempting to update. The client will verify the
+snapshot Merkle metadata file by reconstructing the Merkle tree and comparing
+the computed root hash to the hash provided in timestamp metadata. If the
+hashes do not match, the snapshot Merkle metadata is invalid. Otherwise, the
+client will use the version information in the verified snapshot Merkle
+metadata to proceed with the update.
+
+For additional rollback protection, the client may download previous versions
```
joshuagl, line 154: Is it possible to do this with the PoC for the reference implementation? The files in tests/repository_data/repository/metadata are of the fixed form METADATA_ROLENAME-snapshot.json. How do we identify a previous version to download?

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,276 @@
...
+## Auditing Merkle trees
+
+In order to ensure the validity of all targets metadata version information in the
+Merkle tree, third-party auditors should validate the entire tree each time it
+is updated. Auditors should download every snapshot Merkle file, verify the
+paths, check the root hash against the hash provided in timestamp metadata,
+and ensure that the version information has not decreased for each leaf.
+Alternatively, the repository may provide auditors with information about the
+contents and ordering of leaf nodes so that the auditors can more efficiently
+verify the entire tree.
```
joshuagl, line 171: I'd appreciate some discussion here of what happens if an auditor starts watching a repository that has already gone through several Merkle tree generations. Is it a problem that the auditor hasn't validated all trees generated by the repository?

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,276 @@
...
+## Rollback attack
+
+A rollback attack provides a client with an old, previously valid view of
+the repository. Using this attack, an attacker could convince a client to
+install a version from before a security patch was released.
+
+TUF currently protects against rollback attacks by checking the current time
+signed by timestamp and ensuring that no version information provided by
+snapshot has decreased since the last update. With both of these protections,
+the client is secure against a rollback attack to any version released
```
joshuagl, line 201:
```suggestion
a client that has a copy of trusted metadata is secure against a rollback attack to any version released
```
somewhat implicit in the "before the previous update cycle" below, but possibly worth being explicit

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,276 @@
...
+before the previous update cycle, even if the timestamp and snapshot keys
+are compromised.
+
+Using snapshot Merkle trees, rollback attacks are prevented by both the
+client verification and by third party auditors. If no keys are compromised,
+the timestamp keys protect against a rollback attack by ensuring a valid
+snapshot Merkle tree. If the timestamp key is compromised, the client
+verification of previous Merkle trees provides rollback protection for the
```
joshuagl, line 209: Is this only for a client with previously downloaded Merkle trees? Or are we suggesting that a client could (should?) download previous Merkle trees to ensure the targets metadata file they are downloading isn't susceptible to a rollback attack (I think so based on context below)? In which case, I'd like to see the TAP suggest when the download of previous snapshot Merkle trees might be performed and how.

candidate-snapshot-merkle-tap.md
joshuagl: From context in the security analysis section below, I think an auditor is valid so long as it has the full Merkle tree since the last timestamp key rotation?

candidate-snapshot-merkle-tap.md
mnm678: The current design leaves the hash algorithm to the implementer (to be specified in a POUF). Although, it might be nice to include the algorithm in the timestamp metadata along with the root hash so that the algorithm could change over time without needing a major version release.

Switching from a Merkle tree to something like a Verkle tree might be trickier without a major version release as it changes the whole algorithm for computation rather than just the hash algorithm used.

candidate-snapshot-merkle-tap.md
```diff
@@ -0,0 +1,276 @@
...
+files containing this leaf information and a path to the root of the Merkle tree.
+Clients can then compare this information with a Merkle root provided in the
+timestamp metadata. To prove that there has not been a reversion of the
+snapshot Merkle tree when downloading an image, the client and third-party
+auditors download the prior snapshot Merkle trees and check that the version
+numbers did not decrease at any point.
+
+By replacing a single snapshot metadata file with individual snapshot Merkle
+metadata files, this TAP reduces the metadata overhead for repositories with
+large numbers of targets metadata files.
```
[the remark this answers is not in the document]
mnm678: Good idea, I added a sentence below

candidate-snapshot-merkle-tap.md
mnm678: This isn't possible in the current implementation. I can update the implementation to use consistent snapshots to access previous Merkle trees.

candidate-snapshot-merkle-tap.md
mnm678: I added a paragraph about fast forward attack recovery that describes which versions should be verified by new auditors

candidate-snapshot-merkle-tap.md
mnm678: I added a description of the client previous tree verification to the `Merkle tree verification` section.

mnm678:
Thanks for the review @joshuagl! I responded to your comments inline.

candidate-snapshot-merkle-tap.md
mnm678: I added consistent snapshots to all Merkle tree files in the [PoC](https://github.com/theupdateframework/tuf/pull/1113/commits/342b58fbf661f351cd486d5e098a18529f709093).

joshuagl (commented):
Apologies for taking so long to get back to this PR @mnm678, I've reviewed it this week with fresh eyes. I really like this idea, I think we would benefit from a few additional steps before we merge as a draft:
* minor edits of language for clarity, suggested in-line
* summarising each subsection of the security analysis with a concluding statement comparing snapshot Merkel trees, with and without an auditor, to the current status quo
* better quantifying how much less data clients will need to download under this model
* assign the TAP a number

Before the TAP comes out of draft, I think it would be good to also add:
* PoC implementation of an auditor
* PoC and recommendations for integrating an auditor into the client workflow

mnm678 (commented):
Thanks for the review @joshuagl!

mnm678:
@joshuagl This is ready for another review. I addressed everything except the auditor implementation in the PoC, which I will do before this moves out of the draft stage.

mnm678:
I added a [PoC auditor implementation](https://github.com/theupdateframework/tuf/pull/1113/commits/ec82eb802addec550defa321c1bed33b5fe4ffd8). It doesn't yet include client/auditor interaction, but it demonstrates how an auditor could download and verify all snapshot merkle metadata.

joshuagl (commented):
Thank you for the continued iteration on this TAP Marina! I think this looks in a good shape to merge as a draft to the TAPs repository.

I appreciate you sharing a diff of the reference POUF. However, I don't think we should be merging that until the reference implementation is updated, could you remove that commit from this PR?
Aside: the updated POUF doesn't quite answer the questions I was expecting, either. From the TAP:
> The exact algorithm for generating this Merkle tree (ie the order of nodes in the hash, how version information is encoded, etc.), is left to the implementer, but this
 algorithm should be documented in a POUF

I currently need to read the PoC code and the POUF to understand how to generate the Merkle tree.

---

Following are some thoughts on considerations/work items once the PR is merged:
* A reader has to understand Merkle trees to implement this TAP, for example to understand how to compute the root hash from a leaf node. Is that a concern? Should we link to a good overview of Merkle trees? Should we specify how this is expected to work?
  * are the metadata extensions for the Merkle tree truly abstract enough to support arbitrary (non-Merkle) tree algorithms? Should we PoC some other algorithms? perhaps in a different implementation?
* Further PoC development of auditor integrations?
* Recommendations on specifying the algorithms for a POUF
* The Specification section suggests the snapshot Merkle tree replaces the single snapshot metadata file – is there any reason we shouldn't generate both? If we generate a Merkle tree _only_, should integrations still have a snapshot role with associated key? Should we explore this a bit more?
* In places the TAP seems to suggest an auditor is required, whereas in others it indicates auditors are optional. Let's be sure to clarify.

mnm678:
Thanks again for the review @joshuagl!

> I appreciate you sharing a diff of the reference POUF. However, I don't think we should be merging that until the reference implementation is updated, could you remove that commit from this PR?
> Aside: the updated POUF doesn't quite answer the questions I was expecting, either. From the TAP:
> 
> > The exact algorithm for generating this Merkle tree (ie the order of nodes in the hash, how version information is encoded, etc.), is left to the implementer, but this
> > algorithm should be documented in a POUF
> 
> I currently need to read the PoC code and the POUF to understand how to generate the Merkle tree.

I moved the POUF changes to #133. I'll continue to iterate on this there.

> 
> Following are some thoughts on considerations/work items once the PR is merged:
> 

I moved these to #134 for further discussion.

joshuagl (approved):
Thank you for the continued iteration on this TAP Marina. I think this is ready for merging as a draft.

Issue #121, opened 2020-07-08
TAP process (TAP 1)


joshuagl:
I think the TAP process could use some clarification/modernisation. The following is a list of questions that arose whilst reviewing TAP 1.

* Per TAP 1 the Final state is when the features of a TAP are integrated/implemented within the reference implementation. There is no mention of integration into the specification. Should there be an additional state for "in the spec" vs "in the reference implementation"? Or should Final be "in the spec"?
* Should TAP formats recommend line-wrapping? The existing TAPs mostly seem to line wrap at ~72-79 characters.
* Is Post History, and the mailing-list centric nature of TAP 1, still relevant? It seems to me that it makes sense to discuss TAPS here, on GitHub, with posts to the mailing list to draw interested parties into the discussion? TAP 1 suggests that TAP submissions, reviews and ownership transfers should happen on the mailing list. I haven't seen that happen in practice.
* TAP 1 suggests that 
  > If there are multiple authors, each should be on a separate line following RFC 2822 continuation line conventions.

  that doesn't appear to be true in any of the current TAPs. Should we update the TAPs or change the guidance in TAP 1?
* A TAP is accepted as a Draft, but there's also an Accepted status. It's not clear to me from reading TAP 1 when we are talking about accepted as Draft and accepted as Accepted. The statuses and how a TAP moves through them should be cleared up.

joshuagl:
> * Per TAP 1 the Final state is when the features of a TAP are integrated/implemented within the reference implementation. There is no mention of integration into the specification. Should there be an additional state for "in the spec" vs "in the reference implementation"? Or should Final be "in the spec"?

I absolutely agree with the rationale that the complexity of a TAP may not become clear until someone has tried to implement it, and therefore agree that no TAP should be accepted (Accepted?) until there's a reference implementation. 

My impression has been that what makes a TAP Final is its inclusion in the specification, as that is the TUF artefact that more people are using and engaging with.

My current thought is that a TAP should not go from Draft->Accepted until there's an implementation, but that a TAP moves from Accepted->Final once it's integrated into the specification.

Merging into the reference implementation should follow swiftly, so long as we care about maintaining compatibility with the specification.

trishankatdatadog:
Thanks for raising these questions, Joshua! One concern of mine in particular is whether we should (or can we?) updated TAP 1 to include RFC 2119 language...

joshuagl:
[...]

PR #128, opened 2020-11-04
Add TAP for client initialization and metadata backstop

erickt:
This TAP proposes extending the spec to describe how clients should initialize with local metadata, as well as allowing for a metadata backstop to protect against local rollback attacks.

This grew out of a proposal in https://github.com/theupdateframework/specification/issues/107.

### DIFF
--- a/tap15.md
+++ b/tap15.md
@@ -1,0 +1,477 @@
+* TAP: TBD
+* Title: TUF Metadata Initialization and Backstop
+* Version: 1.0.0
+* Last-Modified: 04-11-2020
+* Author: Erick Tryzelaar
+* Status: Draft
+* Content-Type: text/markdown
+* Requires: <TAP numbers>
+* Created: 18-September-2020
+* TUF-Version: 1.0.10
+* Post-History: <dates of postings to the TUF mailing list>
+
+# Abstract
+
+This TAP describes how a client should initialize a client with local metadata,
+and impose a metadata backstop to limit the range of a potential rollback
+attack during client initialization.
+
+# Motivation
+
+One of the attacks [TUF] protects against is a rollback attack, where an
+attacker tricks a client into using older metadata in order to install a
+vulnerable target. TUF protects against this by requiring all new metadata must
+have a version number greater than the current version. However, TUF is still
+vulnerable to two other avenues to trigger a rollback:
+
+* An initializing client may not have any trusted metadata, so a
+  man-in-the-middle attacker could serve older versions of metadata that
+  include a vulnerable target.
+* A local attacker could replace the local trusted metadata with older versions
+  of metadata.
+
+TUF provides some protection of this rollback during initialization:
+
+* Clients are initialized with a trusted root metadata. This limits a
+  man-in-the-middle to only being able to trick an initializing client into
+  installing metadata that was signed by the keys in the trusted root.
+* Checking for metadata expiration limits a man-in-the-middle to only being
+  able to trick an initializing client into installing unexpired metadata.
+
+Unfortunately these have limitations:
+
+* Rotating the keys in the root role can be a complicated manual process, so
+  it’s expected to be infrequently performed. This can give a man-in-the-middle
+  a serve a wide range of old metadata.
+* Offline devices could use TUF to install updates from some physical medium,
+  like a USB thumb drive. In these circumstances, the TUF metadata could not
+  use a short expiration window in order to account for the time difference
+  from capturing the metadata to when it was used to install packages.
+* A local attacker could both of these protections by modifying the local copy
+  of the trusted metadata, or manipulate the system clock to allow installation
+  of expired metadata.
+
+# Rationale
+
+Some devices and operating systems can establish a trusted boot mechanism to
+establish a chain of trust from a hardware-protected root of trust, which
+verifies the bootloader, which in turn verifies the kernel, and finally the
+user space. For example, UEFI [Secure Boot], Android’s [Verified Boot].  This
+technology can be used to establish a metadata backstop that cannot be
+manipulated by a local attacker, and can be easier to be kept up to date than
+rolling root metadata keys.
+
+## Including all metadata in the backstop
+
+The most direct approach is to bake the backstop metadata into a signed
+read-only partition. A TUF client can initialize trust first with the backstop
+metadata, before loading the local cached metadata, and finally updating with
+the latest remote metadata. Since the backstop is in a signed partition, a
+local attacker cannot replace this metadata with older compromised metadata.
+However, if the repository is large, it can be expensive to bake in the full
+backtop metadata just to prevent a rollback attack.
+
+## Root and Snapshot metadata backstop
+
+Instead it would be sufficient to bake just the Root and Snapshot metadata into
+the backstop. This would prevent a Timestamp rollback by 5.2.2.2, where a new
+Timestamp is rejected if it attempts to roll back the Snapshot version. This
+would prevent a Targets and Delegated Targets rollback by 5.3.3.2, where a new
+Snapshot is rejected if it attempts to roll back the Targets or any Delegated
+Targets metadata.
+
+This alleviates the need to store the Timestamp, Targets, and any Delegated
+Targets, but it still takes up space in the signed partition. On a small
+device, that space may be limited, so it may be too expensive.
+
+However, since it is optional for the Snapshot to contain hashes for the
+Targets and Delegated Targets. This leaves susceptible to the attack detailed
+in [Mercury Paper][Mercury Paper 6.4], where an attacker who has stolen the
+Targets or Delegated Targets keys can forge a Targets role that shares a
+version with the trusted Targets, but points at a compromised target.
+
+## Backstop versions (and optionally, sizes and hashes)
+
+In cases where space in the signed partition is at a premium, the minimum
+needed to implement a backstop is to store in the backstop:
+
+* Root: Root public keys, version, and optionally size in bytes and hashes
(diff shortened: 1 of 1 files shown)


tap15.md
```diff
@@ -0,0 +1,477 @@
...
+  man-in-the-middle to only being able to trick an initializing client into
+  installing metadata that was signed by the keys in the trusted root.
+* Checking for metadata expiration limits a man-in-the-middle to only being
+  able to trick an initializing client into installing unexpired metadata.
+
+Unfortunately these have limitations:
+
+* Rotating the keys in the root role can be a complicated manual process, so
+  it’s expected to be infrequently performed. This can give a man-in-the-middle
+  a serve a wide range of old metadata.
```
trishankatdatadog, line 45: Grammatical error on this line...

jku:
This is very interesting, thanks for taking a stab at it. I'll leave some comments based on a first read -- My POV here is thinking about similar issues with pip where I imagine multiple pip instance that may have different bootstrap metadata (some of them relatively safe, some of them not) but share the same metadata cache. 

Some of the comments could be misguided because I had a bit of trouble understanding which source of metadata is referred to in cases: e.g. steps 5.0 - 5.2 talk about loading _initial root metadata_, then loading _backstop metadata_, loading _root metadata role_ and finally mentions _trusted root metadata_ in a way that is not crystal clear to me.

For the rationale section, I would also mention the simplest case: the original out-of-band metadata may be protected by filesystem permissions or ACLs in a way that makes it more trustworthy than the cached runtime metadata -- no hardware protection is needed to make this a reasonable idea.

My impression is that in the TAP there's a root update loop that verifies the chain of trust from backstop root to most recent cached root (step 5.1), and then later a root update loop from current cached root to current remote repository root (step 5.7). Is there a reason not to consider the backstop root as trusted in the beginning of the process, and do a single *Update the root metadata file*-loop using both the local metadata cache and the remote repository as potential sources for new root versions (let's assume client caches all root versions)?

At first read the process feels complex: for the non-root cases like timestamp, is it not possible to first establish the current "trusted local metadata" (either backstop or cached or nothing based on what's available) and then do the updates from remote repository using the current update process -- making the backstop as much an optional preliminary step to the update as possible? As an example, isn't there redundancy in e.g. the rollback checks or do I misunderstand something: if you make a rollback check for cached timestamp (against backstop timestamp, step 5.3.3) and later a rollback check for downloaded timestamp (against cached timestamp, step 5.8.3), do you really also need to do a rollback check for the downloaded timestamp (against backstop timestamp, step 5.8.2)? 

There are multiple references to "step 5.6." but that does not seem to exist.

5.2.7 Seems to add a root freeze test that seems harsh: update cycle MUST be aborted if the cached root is expired. That is probably not reasonable for some use cases.

tap15.md
```diff
@@ -0,0 +1,477 @@
...
+# Specification
+
+This proposes changing the TUF specification in order to enable clients to
+integrate into one of these trusted boot mechanisms.
+
+(Note: all proposed changes to the TUF spec are in _italics_):
+
+...
+
+_**5.0**. **Load the initial root metadata file.** We assume that a good, trusted
```
mnm678, line 130: The proposed spec changes overall look great, but would benefit from theupdateframework/specification#121 in the future to remove some duplication.

tap15.md
```diff
@@ -0,0 +1,477 @@
...
+    delegated targets files, in the trusted snapshot metadata file equals its
+    version number in the new snapshot metadata file, then the trusted targets
+    metadata hashes, and any delegated targets metadata hashes, if any, MUST
+    equal the hashes in the new snapshot metadata file. If not, discard the new
+    snapshot metadata file, abort the update cycle, and report the potential
+    rollback attack._
+
+*   ...
+
+# Security Analysis
```
mnm678, line 423: It might be worth mentioning a fast forward attack on the backstop snapshot metadata (ie backstop metadata with very high version numbers). It looks like this is mitigated by deleting fast-forwarded snapshot metadata in section 5.7.10, but an explicit mention here might be helpful.

mnm678 (commented):
This looks great! I left a couple of comments

joshuagl (commented):
Thanks for writing this up @erickt. Backstop metadata is a great idea.

It took me a couple of attempts to realise what was going on in the rationale section, where the TAP proposes several backstop approaches. As these will likely be the canonical source for how to implement backstop for clients, it would be good to clean those up such that they are easier to reference. I think it would be good to break them out into their own section, number the approaches, and clearly list the pros and cons of each approach.

As @mnm678 already suggested, I think it would be good to resolve #121 first so that we can simplify the specification section of this TAP. Not least of all because the specification has changed enough that the TAP specification section is out of date.

joshuagl:
> Some of the comments could be misguided because I had a bit of trouble understanding which source of metadata is referred to in cases: e.g. steps 5.0 - 5.2 talk about loading _initial root metadata_, then loading _backstop metadata_, loading _root metadata role_ and finally mentions _trusted root metadata_ in a way that is not crystal clear to me.

Nice observation, it might be good to have a definitions section of the TAP? Then ensure we're being consistent in how we use terms throughout the TAP.

> For the rationale section, I would also mention the simplest case: the original out-of-band metadata may be protected by filesystem permissions or ACLs in a way that makes it more trustworthy than the cached runtime metadata -- no hardware protection is needed to make this TAP a reasonable idea.

💯 

> My impression is that in the TAP there's a root update loop that verifies the chain of trust from backstop root to most recent cached root (step 5.1), and then later a root update loop from current cached root to current remote repository root (step 5.7). Is there a reason not to consider the backstop root as trusted in the beginning of the process, and do a single _Update the root metadata file_-loop using both the local metadata cache and the remote repository as potential sources for new root versions (let's assume client caches all root versions)?

This question made me realise I'm making some assumptions about backstop vs. local cache that we might want the TAP to clarify. 

I don't know if it's what you are suggesting, but I am very wary of the idea of reaching out to the remote repository to fill in "gaps" in the backstop metadata while loading the backstop. It feels like having distinct "load backstop" and "fetch new" stages help keep rollback and freeze attack protection strong. Especially when using backstop metadata means having large expiration times on timestamp and snapshot.

> At first read the process feels complex: for the non-root cases like timestamp, is it not possible to first establish the current "trusted local metadata" (either backstop or cached or nothing based on what's available) and then do the updates from remote repository using the current update process -- making the backstop as much an optional preliminary step to the update as possible? As an example, isn't there redundancy in e.g. the rollback checks or do I misunderstand something: if you make a rollback check for cached timestamp (against backstop timestamp, step 5.3.3) and later a rollback check for downloaded timestamp (against cached timestamp, step 5.8.3), do you really also need to do a rollback check for the downloaded timestamp (against backstop timestamp, step 5.8.2)?

Indeed, this is more like what I'd imagined too. 

> 5.2.7 Seems to add a root freeze test that seems harsh: update cycle MUST be aborted if the cached root is expired. That is probably not reasonable for some use cases.

Well spotted. I wouldn't expect the backstop root having expired to be an issue, much as the bootstrap root expiration doesn't matter in the current client workflow.

Nice review @jku, thanks for taking the time to do it.

tap15.md
```diff
@@ -0,0 +1,477 @@
...
+This TAP describes how a client should initialize a client with local metadata,
+and impose a metadata backstop to limit the range of a potential rollback
+attack during client initialization.
+
+# Motivation
+
+One of the attacks [TUF] protects against is a rollback attack, where an
+attacker tricks a client into using older metadata in order to install a
+vulnerable target. TUF protects against this by requiring all new metadata must
+have a version number greater than the current version. However, TUF is still
```
MVrachev, line 24:
```suggestion
have a version number greater or equal to the current version. However, TUF is still
```

MVrachev:
For now,  I want to focus on the root backstop process and comment it.
1. On step `5.1` : what happens if the loading of backstop metadata versions and hashes fail? Probably end the whole process?
What if it fails only for a certain role? 

2. On step `5.2.2`: `Try loading version N+1 of the root metadata file from the local repository` when we say `local repository` it's good to leave a sentence somewhere explaining that `local` could mean any secure read-only storage you decide to use. In. a perfect scenario you will not access the internet when loading any backstop metadata files, but should we enforce it?
Probably, the best place to clarify this will be a `definitions` section as @joshuagl mentioned.

3. I agree with @joshuagl and I am too worried about the idea for larger expiration times for backstop files. This could easily backfire in multiple ways and it adds its own complexity and limitations for tuf users. Some of the points I can think of:
- If we assume the backstop files have a longer expiration date, then doesn’t this mean that every file in the backstop chain should have a longer expiration time?
We check for freeze attack when we have reached the maximum backstop root version locally available, but this means it could be any version.
- what do we get by checking for a `freeze` attack when the expiration time is made longer?
- if we give larger than usual expiration times for the backstop metadata files, then the user of tuf should be diligent and careful which version of the metadata file he/she uses as backstop.

4. On step 5.2.8: when you say  `delete... and the backstop metadata versions and hashes` from where?

JustinCappos:
For what it is worth with respect to freeze attacks / metadata validity,
the TUF spec was written with the expectation that clients would update and
replace metadata within the lifetime of expiration in the normal case.  So,
I would think it may be reasonable for backstop metadata to have its
expiration information ignored.  (If we wanted to be slightly more
paranoid, we could check to see that some set of it was valid at all times,
but this is an attacker able to replace files in the backstop allowed in
the threat model?)

joshuagl:
Following the discussion at the TUF working session in Paris I've tried to update this PR to address review comments and make it mergeable as a Deferred TAP.

mnm678 (approved):

Issue #126, opened 2020-10-09
Allowing TAPs that are licensed with MIT or Apache 2.0?


erickt:
Good morning. The [TAP-2](https://github.com/theupdateframework/taps/blob/master/tap2.md) template states that TAPs must be placed in public domain, or licensed with [Open Publication License](https://opencontent.org/openpub/). Would you also consider allowing for TAPs that are licensed with MIT or Apache 2.0?

erickt:
Note that [Rust](https://github.com/rust-lang/rfcs#contributions) dual licenses their RFCs with MIT and Apache 2.0, and [Swift](https://github.com/apple/swift-evolution) and [Kubernetes](https://github.com/kubernetes/enhancements) license their extensions with Apache 2.0.

trishankatdatadog:
A few questions:
1. What are the pros and cons of dual-licensing?
2. Would this apply retrospectively? Presumably not, and that is probably fine.

erickt:
1. Unfortunately I am not qualified to answer on the pros and cons of dual licensing. Does CNCF has resources to help answer that kind of question?
2. For my purposes, I wouldn't needed the change to be retroactive. I would just prefer to use one or the other to post new TAPs.

mnm678:
cc @JustinCappos

erickt:
@trishankatdatadog I don't currently have approval to submit changes under Open Publication License, but I do with MIT or Apache 2.0. It would be convenient for me to avoid having to seek approval from work. Also pre-approved are creative commons licenses like [CC-By](https://creativecommons.org/licenses/by/4.0/) if that's more appropriate.

trishankatdatadog:
Got it, thanks for the explanation, Eric!

JustinCappos:
I'm not savvy enough to know how to answer this.  I think we should ask
folks at the CNCF for advice.  I'm looping in Chris A who hopefully can
help to direct this.

erickt:
@JustinCappos: Good morning! Has there been any word from Chris A?

JustinCappos:
There hasn't.  Adding Amye and pinging again...

caniszczyk:
The CNCF IP policy requires code to be Apache v2.0 and docs under CC-BY:

https://github.com/cncf/foundation/blob/master/charter.md#11-ip-policy

If dual licensing this as Apache v2.0 + CC-BY works for you that's fine (even though it isn't code per se).

joshuagl:
[...]

PR #129, opened 2020-11-30
Update TAP process to recommend public domain or CC-By for TAPs

joshuagl:
Per the discussion in #126, TAPs should be either placed in the public domain _or_ licensed under the Creative Commons Attribution 4.0 International License (CC-By). This PR updates tap1 & tap2 to remove references to the Open Publication License, and instead recommend CC-By for non-public domain TAPs.

Also fixup the comment in the tap1 dot file which suggests how to generate a png of the flow diagram.


trishankatdatadog:
There goes our plans to get rich quick... 😢

mnm678 (approved):

PR #138, opened 2021-04-30
TAP-17: Remove signature wrapper from TUF spec

adityasaky:
TAP-17 proposes dropping the signature wrapper specification (section 4.2 of the TUF spec), and instead presenting some properties that a signature wrapper must have for use in a TUF implementation.

The proposed POUF-1 update changes the signature wrapper from the current one to DSSE v1.0. This is now in #143

---

Original message:

This TAP proposes replacing the existing signature wrapper of TUF with [DSSE](https://github.com/secure-systems-lab/dsse). in-toto, TUF's sister project, has had a similar effort. That proposal, [ITE-5](https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc) was authored by @SantiagoTorres and @MarkLodato.

### DIFF
--- a/tap17.md
+++ b/tap17.md
@@ -1,0 +1,110 @@
+* TAP: 17
+* Title: Remove Signature Wrapper from the TUF Specification
+* Version: 1
+* Last-Modified: 11/11/2021
+* Author: Aditya Sirish A Yelgundhalli, Marina Moore
+* Type: Standardization
+* Status: Draft
+* Content-Type: markdown
+* Created: 30/04/2021
+* Requires: TAP-11, TAP-14
+* +TUF-Version:
+* +Post-History:
+
+# Abstract
+
+This TUF Augmentation Proposal (TAP) proposes removing the definition of a
+specific signature wrapper and key definitions, and instead defines certain
+properties a wrapper must have. Further, it suggests POUF-1 as an example
+implementors can refer to when choosing to generate TUF metadata.
+
+# Specification
+
+The TUF specification as of v1.0.25 uses a custom signature wrapper. At the
+time of authoring this document, the primary reference implementation written
+in Python also generates TUF metadata using the same signature wrapper.
+
+However, TUF does not mandate the use of this signature wrapper, nor any
+specific metaformat. Indeed,
+[TAP-11, "Using POUFs for Interoperability"](https://github.com/theupdateframework/taps/blob/master/tap11.md)
+enables adopters to make their own decisions for their implementations, and
+provides a mechanism for them to document their decisions.
+[POUF-1](/POUFS/reference-POUF/pouf1.md) is the POUF for the official reference
+implementation, and it seems like the obvious choice for this information to be
+specified.
+
+Section 4.2 of the TUF specification, titled "File formats: general principles"
+may be replaced by a description of the properties that any signature wrapper used
+by a TUF implementation must provide. Some important properties:
+
+* SHOULD include an authenticated payload type
+* SHOULD avoid depending on canonicalization for security
+* SHOULD NOT require the verifier to parse the payload before verifying
+* SHOULD NOT require the inclusion of signing key algorithms in the signature
+* MUST support the inclusion of multiple signatures in a file
+* SHOULD support a hint indicating what signing key was used, i.e., a KEYID
+
+The presence of an authenticated payload type can be valuable for a project like TUF,
+with multiple implementations and derivatives. Indeed, every POUF that describes an
+implementation MUST choose a unique payload type, ensuring that there is no confusion
+about which implementation generated some piece of metadata.
+
+# Motivation
+
+TAP-11 introduced the concept of POUFs but the TUF specification continues to
+specify example formats, namely those used by the reference implementation as
+of June 2021. These definitions are essentially replicated in POUF-1, which is
+meant to be the authoritative source for information about the reference
+implementation. By replacing these definitions with *properties* that a wrapper
+must possess, the specification can aid adopters with the development of their
+implementations and the POUF can serve as an example. In this scenario, both
+documents are serving the purpose originally envisioned for them.
+
+Further, the examples used in the specification are from the old signature
+wrapper that includes certain side effects:
+* it requires canonicalization before signature verification
+* it does not allow for distinguishing different implementations that may have
+  slightly different formats, i.e., it's missing a payload type
+
+# Rationale
+
+Moving the signature wrapper details out of the specification, and instead
+requiring adopters to refer to POUFs for specifics of an implementation ensures
+a clean separation between implementation details and the TUF specification.
+Indeed, it also ensures that the focus of the reader is on only the TUF
+primitives rather than burdening them with the specifics of the signature
+wrapper.
+
+# Security Analysis
+
+Any implementations that build on the properties listed in this document
+will have their security enhanced.
+
+# Backwards Compatibility
+
+The changes proposed in this TAP are backwards compatible with respect to the
+TUF specification. However, for implementations looking to switch to a
+signature wrapper with the properties described here, the change may be
+backwards incompatible. In these instances, the implementations SHOULD set a
+transition period during which they support both old-style and new-style
+envelopes. This transition period MUST be clearly communicated to their users
+using their standard channels.
+[TAP-14, "Managing TUF Versions"](https://github.com/theupdateframework/taps/blob/master/tap14.md)
+contains some useful information about distributing metadata in multiple formats
+that can be used during the transition period.
+
+# Augmented Reference Implementation
+
+TODO: POUF-1 will be updated separately, along with the implementation itself.
+See POUF-1 for details about the reference implementation.
+
+# Copyright
+
+This document has been placed in the public domain.
+
+# References
+
+* [TAP-11](https://github.com/theupdateframework/taps/blob/master/tap11.md)
+* [TAP-14](https://github.com/theupdateframework/taps/blob/master/tap14.md)
+* [File formats in TUF Specification](https://theupdateframework.github.io/specification/latest/index.html#file-formats-general-principles)
+* [POUF-1](/POUFS/reference-POUF/pouf1.md)
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
 * [TAP 14: Managing TUF Versions](tap14.md)
 * [TAP 15: Succinct hashed bin delegations](tap15.md)
 * [TAP 16: Snapshot Merkle Trees](tap16.md)
+* [TAP 17: Remove Signature Wrapper from the TUF Specification](tap17.md)
 
 ## Rejected
 


tap17.md
```diff
@@ -0,0 +1,152 @@
...
+*   `message`: the signed message as an implementation-language-specific object
+*   `signers`: set of recognized names that have signed the message
+
+Steps:
+
+*   `envelope` := JsonDecode(`file`); raise error if the decoding fails
+*   If `envelope.payload` exists (new-style envelope):
+    *  If `payloadType` != `application/vnd.tuf+json`, raise error
+    *  `preauthEncoding` := PAE(UTF8(`envelope.payloadType`),
+        `envelope.payload`) as per signing-spec
```
mnm678, line 66: Can you link to this part of the spec?

tap17.md
```diff
@@ -0,0 +1,152 @@
...
+        (`signature`) in `envelope.signatures` and (`name`, `publicKey`) in
+        `recognizedSigners`
+    *  `message` := `envelope.signed`
+*   Else, raise error
+*   Raise error if `signers` is empty
+*   Return `message` and `signers`
+
+# Motivation
+
+TUF's sister project, in-toto, reused the current signature envelope to maximize
```
mnm678, line 85:
```suggestion
TUF's sister project, in-toto, reused TUF's current signature envelope to maximize
```

tap17.md
```diff
@@ -0,0 +1,152 @@
...
+
+At first sight this proposal is central to security, yet the actual
+contribution is to allow for a signature provider to be disaggregated from the
+specification. As such, no software update-specific security properties are
+removed from the system through this TAP.
+
+The adoption of SSL signing spec slightly improves the security stance of
+implementations because they are no longer parsing untrusted input.
+
+# Backwards Compatibility
```
mnm678, line 128: We should probably also make a plan to phase out the old signature envelope.

tap17.md
```diff
@@ -0,0 +1,152 @@
...
+Enhancement (ITE),
+[ITE-5](https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc).
+
+# Specification
+
+The specification adopted will be the SSL Signing Spec 0.1, as linked above. As
+such, we defer to that document to describe the specifics of signature
+generation and verification.
+
+The envelope's `payloadType` is `application/vnd.tuf+json` for all TUF roles.
```
mnm678, line 39: To include support for TUF implementations that don't use json, can you make this a bit more general? Maybe `application/vnd.tuf+EXT`.

mnm678 (commented):
Thanks for all the work on this @adityasaky! I think this will be a great simplification to the spec. My main concern is making sure that this will interact well with [TAP 11](https://github.com/theupdateframework/taps/blob/master/tap11.md), so making it clear that a specific json encoding isn't required, but an implementation could use ASN.1, etc if they do the proper checks.

adityasaky:
Thanks for the review, @mnm678! I've resolved the easier comments. I'll take a look at TAP-11 in the next few days and see if something jumps out as an issue with this one. :)

POUFs/pouf2.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Usage
+
+Refer to POUF-1.
+
+# Formats
+
+## General Principals
+
+All signed metadata objects have the format defined in DSSE v1:
```
adityasaky, line 32: I think we can go one step further and only point to DSSE instead of showing its envelope here. We'd then have to fill in the other file formats, I reckon.

POUFs/reference-POUF/pouf1.md
```diff
@@ -67,19 +67,22 @@ The following steps must be completed before any updates can be installed:
...
 
 ## General Principals
-All signed metadata objects have the format:
-
-       { "signed" : ROLE,
-         "signatures" : [
-            { "keyid" : KEYID,
-              "sig" : SIGNATURE }
-            , ... ]
+All signed metadata use v1 of [Dead Simple Signing Envelope (DSSE)](https://github.com/secure-systems-lab/signing-spec):
```
adityasaky, line 70: TODO: update link to spec after v1 is tagged and repository is renamed.

adityasaky:
I've updated POUF-1 to use DSSE, though there are some TODOs we need to fill in before this can be merged. I've also updated TAP-17 to remove the signature wrapper from TUF entirely, and instead lean entirely on POUFs for that information.

adityasaky:
Thanks a ton, @mnm678, for helping me draft this. :smile:

POUFs/reference-POUF/pouf1.md
```diff
@@ -400,13 +403,79 @@ matching paths will only be tried if downloading from earlier mirrors fails.
 This behavior can be modified by the client code that uses the framework to,
 for example, randomly select from the listed mirrors.
 
+# Transitioning from the old signature wrapper
```
adityasaky, line 406: @trishankatdatadog @mnm678 please let me know if this makes sense.

tap17.md
adityasaky: Resolving this, I think that's now part of the POUF.

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+adopters to make their own decisions for their implementations, and provides a
+mechanism for them to document their decisions. POUF-1 is the POUF for the
+official reference implementation, and it seems like the obvious choice for this
+information to be specified.
+
+Section 4.2 of the TUF specification, titled "File formats: general principles"
+may be replaced by a description of the properties that any signature wrapper used
+by a TUF implementation must provide. Some important properties:
+
+* MUST include an authenticated payload type
```
adityasaky, line 38: Does this list of properties make sense from the perspective of TUF?

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+* SHOULD avoid depending on canonicalization for security
+* SHOULD NOT require the verifier to parse the payload before verifying
+* SHOULD NOT require the inclusion of signing key algorithms in the signature
+* MUST support the inclusion of multiple signatures in a file
+* SHOULD support a hint indicating what signing key was used, i.e., a KEYID
+
+The presence of an authenticated payload type can be valuable for a project like TUF,
```
adityasaky, line 45: I'd like to hear some thoughts here. Should two implementations designed by the same team to be compatible, say a Python implementation and a Go implementation, have different payload types? Or should they have the same one because in essence they're described by the same POUF? Here I'm leaning towards the latter.

mnm678: I think implementations with the same POUF should have the same payload type. That way it can be used to determine if the metadata is compatible with a particular implementation.

adityasaky: Perhaps the proposed payload type for POUF-1 should contain the POUF number somewhere.

erickt (commented):
While I like the idea of using DSSE, I think we need to consider the migration path for old clients that don't understand this new envelope format. So I think it makes more sense to treat this as POUF-2, rather than by overwriting POUF-1. I'm guessing repositories will need to either:

* Create a POUF-1 repository and a POUF-2 repository. Publish a new client to POUF-1 which understands how to read POUF-2, and start fetching from the new repository. Servers would need to make sure the POUF-1 repository's metadata doesn't expire until the old client versions no longer receive support.
* Create both a POUF-1 and POUF-2 repositories, and publish updates to both repositories until the old clients no longer receive support.

POUFs/reference-POUF/pouf1.md
```diff
@@ -1,11 +1,11 @@
 * POUF: 1
```
erickt, line 1: Could you explain why you defined this in POUF-1, rather than creating a new POUF? We would need to go through a transition period, and so we should still say the old signature format was valid. Perhaps we could just mark POUF-1 as deprecated instead?

POUFs/reference-POUF/pouf1.md
```diff
@@ -1,11 +1,11 @@
...
-* Title: Reference Implementation Using Canonical JSON
-* Version: 2
-* Last-Modified: 06-May-2020
+* Title: Reference Implementation Using Canonical JSON and DSSE
```
erickt, line 2: If we're making a backwards incompatible change to the format, it'd be really nice if we could address the whole [canonical json is not json](https://github.com/theupdateframework/specification/issues/92) issue.

POUFs/reference-POUF/pouf1.md
adityasaky: This was originally [defined](https://github.com/theupdateframework/taps/pull/138/commits/4cbb8d5d2002caffb46f023358d27a384821d5c3) as a separate POUF, but then moved to POUF-1 because it's still pertinent to the reference implementation. Perhaps it's worth cloning the current POUF-1 and releasing a new version as per https://github.com/theupdateframework/taps/blob/master/tap11.md#pouf-storage.

cc @mnm678

POUFs/reference-POUF/pouf1.md
erickt: Actually, would we still need canonicalization if we used this envelope format? I don’t think the update flow would ever need to re-serialize the metadata after verification.

POUFs/reference-POUF/pouf1.md
mnm678: I advocated for keeping this in POUF 1 in order to make it clear that this is still the POUF for the reference implementation. We'd probably wait to merge the changes to POUF 1 until the reference implementation matches the state of the POUF. However, if that is adding confusion, I'd be alright with copying this content into a new file.

mnm678 (approved):

trishankatdatadog (approved):

POUFs/reference-POUF/pouf1.md
```diff
@@ -14,7 +14,7 @@ This POUF describes the protocol, operations, usage, and formats for the TUF ref
 
 The reference implementation includes all required features of the TUF standard, as well as many of the optional features as a reference for anyone wishing to implement TUF. The implementation uses Canonical JSON encoding.
 
-This version of the POUF covers v0.12.* of the reference implementation and has been updated to reflect that: snapshot.json only lists targets metadata (top-level and delegated), and timestamp.json includes hashes and length in METAFILES.
+This version of the POUF covers v0.12.* of the reference implementation and has been updated to reflect that: snapshot.json only lists targets metadata (top-level and delegated), and timestamp.json includes hashes and length in METAFILES. TODO: update this bit
```
joshuagl, line 17: I don't think we should be updating the POUF until the reference implementation implements TAP, and this TODO reinforces that.

tap17.md
```diff
@@ -0,0 +1,94 @@
+* TAP: 17
+* Title: Remove Signature Wrapper from the TUF Specification
+* Version: 0
```
joshuagl, line 3:
```suggestion
* Version: 1
```

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+* Last-Modified: 22/06/2021
+* Author: Aditya Sirish A Yelgundhalli, Marina Moore
+* Type: Standardization
+* Status: Draft
+* Content-Type: markdown
+* Created: 30/04/2021
+* +TUF-Version:
```
joshuagl, line 10:
```suggestion
* Requires: TAP 14
* +TUF-Version:
```
I think this makes sense, because we won't be merging any changes to the spec related to this TAP until TAP 14 has been accepted.

POUFs/reference-POUF/pouf1.md
```diff
@@ -400,13 +403,79 @@ matching paths will only be tried if downloading from earlier mirrors fails.
...
+    *  `message` := `envelope.signed`
+*   Else, raise error
+*   Raise error if `signers` is empty
+*   Return `message` and `signers`
+
+Note: This workflow was authored by Mark Lodato (@MarkLodato) for in-toto.
 
 # Security Audit
-This profile was included in TUF security audits available at https://theupdateframework.github.io/audits.html.
+This profile was included in TUF security audits available at https://theupdateframework.github.io/audits.html. TODO: the new wrapper has not been audited
```
joshuagl, line 456: Another TODO which suggests the POUF update should come after the reference implementation has been changed.

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+# Abstract
+
+This TUF Augmentation Proposal (TAP) proposes removing the definition of a
+specific signature wrapper and key definitions, and instead defines certain
+properties a wrapper must have. Further, it suggests POUF-1 as an example
+implementors can refer to when choosing to generate TUF metadata.
+
+# Specification
+
+The TUF specification currently, as of June 2021, uses a custom signature
```
joshuagl, line 22:
```suggestion
The TUF specification currently, as of June 2021, uses a custom signature
```
I wonder if we should mention versions, rather than dates, here? TUF uses semver, so we should be able to reference the major version.

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+wrapper. At the time of authoring this document, the primary reference
+implementation written in Python also generates TUF metadata using the same
+signature wrapper.
+
+However, TUF does not mandate the use of this signature wrapper, nor any
+specific metaformat. Indeed, TAP-11, "Using POUFs for Interoperability" enables
```
joshuagl, line 28: Nit: link to TAP 11 here might be nice

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+adopters to make their own decisions for their implementations, and provides a
+mechanism for them to document their decisions. POUF-1 is the POUF for the
```
joshuagl, line 30: Nit: link to POUF 1 here might be nice

tap17.md
joshuagl: Agree, we want the payload type to describe the format not the implementation. If go-tuf and python-tuf want to have compatible implementations, they should use the same payload type.

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+meant to be the authoritative source for information about the reference
+implementation. By replacing these definitions with *properties* that a wrapper
+must possess, the specification can aid adopters with the development of their
+implementations and the POUF can serve as an example. In this scenario, both
+documents are serving the purpose originally envisioned for them.
+
+Further, the examples used in the specification are from the old signature
+wrapper that includes certain side effects:
+* it requires canonicalization before signature verification
+* it does not allow for distinguishing different implementations that may have slightly different formats, i.e., it's missing a payload type
```
joshuagl, line 64: Nit: could you wrap this line?

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Rationale
+
+Moving the signature wrapper details out of the specification, and instead
+requiring adopters to refer to POUFs for examples ensures a clean separation
```
joshuagl, line 69:
```suggestion
requiring adopters to refer to POUFs for specifics of an implementation ensures a clean separation
```

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+than burdening them with the specifics of the signature wrapper.
+
+# Security Analysis
+
+Any implementations that build on the properties listed in this document
+will have their security enhanced.
+
+# Backwards Compatibility
+
+No backwards incompatibility.
```
joshuagl, line 81: I think this section would benefit from some more content. It's strictly true that it's not backwards incompatible for the specification, but the specification has a symbiotic relationship with the reference implementation. A change of the signature wrapper in the reference implementation absolutely is backwards incompatible, as the POUF update indicates.

tap17.md
```diff
@@ -0,0 +1,94 @@
...
+# Augmented Reference Implementation
+
+See POUF-1 for details about the reference implementation.
```
joshuagl, line 84: Let's mark this TODO?

joshuagl (commented):
Thanks for submitting this TAP @adityasaky ! 

In principle, I am on-board with the TAP. I'd like to see a few changes before we merge it as draft, though:
* drop edits to the POUF until such a time as the reference implementation is updated _or_ make this a separate POUF as suggested in https://github.com/theupdateframework/taps/pull/138#discussion_r658812047
* Some discussion in the TAP itself about updating an implementation between wire formats and requiring TAP 14

POUFs/reference-POUF/pouf1.md
adityasaky: Seems reasonable. Perhaps I submit this TAP in this PR which we can go ahead and merge and the changes to the POUF in a separate PR that we leave unmerged for now?

tap17.md
trishankatdatadog: Yeah, where are the POUFs anyway? Are they shared between TUF and Uptane?

POUFs/reference-POUF/pouf1.md
adityasaky: I'm submitting POUF-1 changes in a separate PR which can be merged later, once the reference implementation is ready to make the switch.

adityasaky: #143 has it now.

adityasaky:
I believe I've addressed everything other than the backward compatibility comment, @joshuagl. Thank  you for the review!

On that front: perhaps we can discuss it in the context of it being a backward incompatible change for anyone switching from the current wrapper to any other wrapper, DSSE or otherwise, and therefore they must provide a transition period and clear communication of the breaking change? This is what we have in ITE-5. See: https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc#backwards-compatibility

POUFs/reference-POUF/pouf1.md
adityasaky: Sorry @erickt, I missed this. Canonicalization shouldn't be necessary when using DSSE. One of its [goals](https://github.com/secure-systems-lab/dsse/blob/master/background.md) is to avoid canonicalization, therefore ensuring possibly untrusted inputs are not parsed. When the implementation makes the switch and stops relying on canonicalization, I suspect https://github.com/theupdateframework/specification/issues/92 should be addressed, but perhaps someone more qualified than me can weigh in on that.

tap17.md
erickt: Yeah, we are quite interested in switching to DSSE, and we’d appreciate any guidance on making the transition. Our situation is complicated by us having to support products that don’t support DSSE need to use TUF to update to the latest version.

I’m guessing we will have to either use a tombstone repo which updates us to some new version that points at the new DSSE repo, or we start publishing the metadata in both formats, where the metadata files have different file extensions.

mnm678: @erickt I wrote a [TAP](https://github.com/theupdateframework/taps/blob/master/tap14.md) about managing TUF versions that describes a version of publishing metadata using multiple formats that you might be interested in. I'm hoping we can get that or something similar working before adding this and other breaking changes to the spec.

adityasaky:
[ITE-5](https://github.com/in-toto/ITE/blob/master/ITE/5/README.adoc), which describes essentially the same change for in-toto, was accepted today!

tap17.md
adityasaky: I just pushed some changes discussing a transition period (on a per implementation basis) during which metadata is published in multiple formats. Also links to TAP-14.

adityasaky:
With the changes to backwards compatibility (https://github.com/theupdateframework/taps/pull/138#discussion_r747806562), I wonder if we can merge this TAP as a draft? Should it presented at the community meeting next week first?

tap17.md
```diff
@@ -0,0 +1,110 @@
+* TAP: 17
+* Title: Remove Signature Wrapper from the TUF Specification
+* Version: 1
+* Last-Modified: 11/11/2021
```
adityasaky, line 4: hah, no date format confusions!

mnm678:
> With the changes to backwards compatibility ([#138 (comment)](https://github.com/theupdateframework/taps/pull/138#discussion_r747806562)), I wonder if we can merge this TAP as a draft? Should it presented at the community meeting next week first?

It's probably ready to merge, but we can present it at the community meeting to be sure.

tap17.md
```diff
@@ -0,0 +1,110 @@
...
+* SHOULD avoid depending on canonicalization for security
+* SHOULD NOT require the verifier to parse the payload before verifying
+* SHOULD NOT require the inclusion of signing key algorithms in the signature
+* MUST support the inclusion of multiple signatures in a file
+* SHOULD support a hint indicating what signing key was used, i.e., a KEYID
+
+The presence of an authenticated payload type can be valuable for a project like TUF,
+with multiple implementations and derivatives. Indeed, every POUF that describes an
+implementation MUST choose a unique payload type, ensuring that there is no confusion
+about which implementation generated some piece of metadata.
```
joshuagl, line 50: Should the POUF format be updated to include capturing the unique payload type for an implementation? Should we provide any guidance on forming payload type?

The spec should probably more strongly recommend capturing the implementation details in a POUF, especially the payload type.

joshuagl (approved):
Thanks for this TAP @adityasaky! It LGTM to merge as a draft. I think questions to answer are:
* Should the spec recommend/suggest/mention a format which implements the proposed properties? i.e., DSSE
* what should the pedagogical examples look like after the spec is updated per this TAP? do we continue to use JSON for file format examples?
* should we include guidance on payload type, particularly given this is a motivation for the TAP?
* what recommendations should the spec make about capturing implementation details in a POUF, especially payload type (how to identify the implementation)?

I've captured these questions in a discussion thread https://github.com/theupdateframework/taps/issues/144

tap17.md
adityasaky: > Should the POUF format be updated to include capturing the unique payload type for an implementation?

This is probably a good idea. @mnm678 what do you think?

TAP-11 lists only the minimum fields a POUF must contain, so in theory it can be extended without changing the TAP, but it's probably worth adding a field. I also wonder if it's worth adding something to formally identify the implementations a POUF describes. POUF-1 identifies the python implementation, but does it also describe the Go implementation, for example? Should it link to them?

adityasaky: > Should we provide any guidance on forming payload type?

I'm not sure we can, since we're only saying the envelope should support an authenticated payload type, while each envelope format may enforce its own standard. Guidance here may run afoul of some signature wrapper that is otherwise compliant? I think the selection of a unique payload type that conforms to a particular wrapper's specification is sufficient, as long as it's recorded in the corresponding POUF.

adityasaky: Maybe it's worth referring to a survey of wrappers to understand their guidelines for payload types. We may be able to specify some aspects like "include information about encoding" without running into trouble.

mnm678: > > Should the POUF format be updated to include capturing the unique payload type for an implementation?
> 
> This is probably a good idea. @mnm678 what do you think?
> 
> TAP-11 lists only the minimum fields a POUF must contain, so in theory it can be extended without changing the TAP, but it's probably worth adding a field. I also wonder if it's worth adding something to formally identify the implementations a POUF describes. POUF-1 identifies the python implementation, but does it also describe the Go implementation, for example? Should it link to them?

I agree, I think we could add payload type as a subsection of Formats in the POUF definition.

I would say the implementations should/could link to the POUF rather than the other way around. The implementation implements TUF plus some POUF.

adityasaky: > I would say the implementations should/could link to the POUF rather than the other way around. The implementation implements TUF plus some POUF.

Makes sense to me.

joshuagl: SGTM, I think we can file an issue against TAP 11 to discuss payload type once we merge this as draft.

joshuagl:
Three TAP editor approvals, merging this as a draft. Thanks for the TAP @adityasaky !

PR #143, opened 2021-09-10
Update reference POUF to use DSSE

adityasaky:
Splitting POUF updates from #138.

cc @joshuagl @mnm678 @erickt @trishankatdatadog


trishankatdatadog:
Ok, but where do we list POUFs? Shouldn't we list it in a README?

mnm678:
> Ok, but where do we list POUFs? Shouldn't we list it in a README?

Good question. We should probably add a readme to this directory.

joshuagl:
I've converted this PR to be a draft as we don't want to merge this until it reflects the reference implementation.

PR #146, opened 2021-11-30
TAP-15: clarify all delegation attributes are optional

joshuagl:
The "succinct_hash_delegations" attribute was missing an or symbol ('|')
between it and the "paths" attribute. This patch re-orders the optional
delegation attributes, sorting them alphabetically (with the happy outcome
that the more complex optional attribute comes last), and ensure's there's
an or symbol ('|') separating each optional attribute.

Signed-off-by: Joshua Lock <jlock@vmware.com>

### DIFF
--- a/tap15.md
+++ b/tap15.md
@@ -177,12 +177,12 @@ With the addition of succinct hashed bins, the delegation will contain:
        ("name": ROLENAME,)
        "keyids" : [ KEYID, ... ] ,
        "threshold" : THRESHOLD,
-       ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
-       ("succinct_hash_delegations" : {
+       ("path_hash_prefixes" : [ HEX_DIGEST, ... ], |
+       "paths" : [ PATHPATTERN, ... ], |
+       "succinct_hash_delegations" : {
          "delegation_hash_prefix_len" : BIT_LENGTH,
          "bin_name_prefix" : NAME_PREFIX
        })
-        "paths" : [ PATHPATTERN, ... ]),
        "terminating": TERMINATING,
    }, ... ]
  }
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

tap15.md
```diff
@@ -177,12 +177,12 @@ With the addition of succinct hashed bins, the delegation will contain:
        ("name": ROLENAME,)
        "keyids" : [ KEYID, ... ] ,
        "threshold" : THRESHOLD,
-       ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
-       ("succinct_hash_delegations" : {
+       ("path_hash_prefixes" : [ HEX_DIGEST, ... ], |
```
trishankatdatadog, line 180: Hmm, this might be accidentally going a bit too far, but not that I think the problem wasn't already there: don't we mean to say that exactly one of these should be specified? What happens if zero or more than one is specified?

mnm678: A couple paragraphs up talks about prioritizing `succinct_hash_delegations` over `path_hash_prefixes`. I don't remember why `paths` wasn't included there

joshuagl: I think we do mean to say only one should be specified. The same problem exists in the [spec today](https://theupdateframework.github.io/specification/v1.0.26/#file-formats-targets), where we say:
> The "path_hash_prefixes" and "paths" attributes are OPTIONAL, if used, exactly one of them should be set.

mnm678: I opened #147 to clarify this in the text

joshuagl:
Any objections to merging this? I don't mean to nag, it's only been 10 days, but I know Marina wants to consider promoting this TAP to Accepted next week.

trishankatdatadog (approved):

PR #148, opened 2021-12-15
Move TAP 15 to Accepted

mnm678:
As discussed in the community meeting, this pr proposes moving TAP 15 to accepted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@
 * [TAP 10: Remove native support for compressed metadata](tap10.md)
 * [TAP 11: Using POUFs for Interoperability](tap11.md)
 * [TAP 12: Improving keyid flexibility](tap12.md)
+* [TAP 15: Succinct hashed bin delegations](tap15.md)
 
 ## Draft
 
 * [TAP 8: Key rotation and explicit self-revocation](tap8.md)
 * [TAP 13: User Selection of the Top-Level Target Files Through Mapping Metadata](tap13.md)
 * [TAP 14: Managing TUF Versions](tap14.md)
-* [TAP 15: Succinct hashed bin delegations](tap15.md)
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
Why SHA2-256? Then again, why do we use that to fix keyids? Might as well fix that everywhere or nowhere. So we can punt this issue.

joshuagl (approved):
I approve marking this TAP as Accepted.

> Why SHA2-256?
That's what we're already using for hashed bin delegations, see `[path_hash_prefixes](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes)`

We really need to better document hashed bin delegation in the spec, should I file an issue on the spec repo?

mnm678:
> We really need to better document hashed bin delegation in the spec, should I file an issue on the spec repo?

Yes, that would be helpful, thanks. I find myself directing people to PEP 458 to explain this rather than the spec.

joshuagl (dismissed):

lukpueh:
Needs rebasing on top of just merged https://github.com/theupdateframework/taps/pull/155. I'd like to submit the discussed follow-up (https://github.com/theupdateframework/taps/pull/155#discussion_r899977971) early next week.

But I'm fine with accepting TAP 15 before that.

joshuagl:
Could we update the "Augmented Reference Implementation" section to point to the merged implementation? https://github.com/theupdateframework/python-tuf/pull/2010

We should also link to some documentation in python-tuf once we have it

lukpueh (changes requested):
@mnm678, did you really rebase? There is a merge commit in your feature branch now. Also, there still is a conflict.

lukpueh (approved):

joshuagl (approved):

PR #149, opened 2022-01-04
TAP 13: Describe interaction with TAP 4

mnm678:
Add a section to TAP 13 that describes the interaction with TAP 4.

I also added the definition for TAP 13 metadata, and added a threshold to the required data.

cc @joshuagl

### DIFF
--- a/tap13.md
+++ b/tap13.md
@@ -172,9 +172,27 @@ cases where the key needs to be rotated or changed and the key is still accessib
 ## Interaction with TAP 4
 
 If a client is using TAP 4 to provide mapping metadata to multiple repositories,
-the TAP 13 targets mapping metadata will be applied to all repositories. This
-means that the targets metadata listed in the targets mapping metadata must
-be present on all repositories listed in TAP 4.
+they MAY provide a TAP 13 targets mapping for each repository.
+An optional `target-mapping` field will be added to TAP 4 to provide this mapping
+when TAP 13 is used. This field will contain:
+
+```
+"target-mapping": {
+   "repositories": [REPOSITORY_NAME, ...],
+   "targets_rolename": ROLENAME,
+   "threshold": THRESHOLD,
+   "keys":{
+      KEYID : KEY,
+      ...
+   }
+}
+```
+
+Where `REPOSITORY_NAME` is the name of the target repository defined in TAP 4,
+and the other fields are as described above.
+
+If a client is not using TAP 4, the targets mapping may instead be in a separate
+metadata file as described above.
 
 # Security Analysis
 
--- a/tap13.md
+++ b/tap13.md
@@ -40,7 +40,7 @@ To allow for safer use of these untrusted repositories, we propose adding
 namespaces to TUF repositories which enable explicit trust decisions. In this
 mode, if Alice and Bob both use repository X and ask for package foo, they may
 get different results based on their trusted namespaces.
-In summary; this proposal enables clients to restrict the targets they consume
+In summary; this proposal enables clients to restrict the targets roles they consume
 to filtered views of the repository.  
 
 These different views could be defined by either different users on the
@@ -70,7 +70,7 @@ specific packages and Alice wishes to install only those packages.  Every other
 user clearly should not be subject to those constraints.
 
 Second, Alice may be concerned that a full repository compromise may include
-the root role.  Since the root role in TUF indicates the top-level target's
+the root role.  Since the root role in TUF indicates the top-level targets'
 role key, this compromise can enable the attacker full control of Alice's
 namespace.  Alice may want to require that the security team at her company
 still be used to decide which packages to trust.  
@@ -136,7 +136,7 @@ to which keys indicate the targets role. The resulting metadata will look like:
 ```
 
 As such, we add to the [Mechanisms that Assigns Targets to Repositories](https://github.com/theupdateframework/taps/blob/master/tap4.md#mechanism-that-assigns-targets-to-repositories)
-support for a reference to the targets file in an identical way to the
+support for a reference to the targets metadata file in an identical way to the
 root file's reference in the [TUF specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats).
 However, additionally, the file name must be specified as this is no longer
 targets.json.
@@ -149,7 +149,7 @@ files.  All targets metadata files (top-level and otherwise) are also stored in 
 same METAPATH location listed in snapshot.json.
 
 The changes in the client application workflow are fairly minor from this
-TAP.  Steps 4.0 and 4.4.0 should refer to the specified target's metadata file instead
+TAP.  Steps 4.0 and 4.4.0 should refer to the specified targets' metadata file instead
 of the top-level targets metadata file.  Additionally, instead of verifying the targets metadata
 file using the key in the root metadata in step 4.0, verification must use the
 keys listed in the mapping metadata.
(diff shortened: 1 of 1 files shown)


tap13.md
```diff
@@ -148,16 +159,22 @@ use mode clearer in the specification.
...
 TAP must take more care.  Thus, one should take into account the operational difficultly to touch
-clients in the case of key loss or compromise for the top-level targets metadata file.  If it is 
-operationally difficult to touch the clients, then the client may perhaps use a threshold of 
-offline keys before delegating to a developer’s key.  [TAP 8](tap8.md) also provides support for 
+clients in the case of key loss or compromise for the top-level targets metadata file.  If it is
+operationally difficult to touch the clients, then the client may perhaps use a threshold of
+offline keys before delegating to a developer’s key.  [TAP 8](tap8.md) also provides support for
 cases where the key needs to be rotated or changed and the key is still accessible to the developer.
 
+## Interaction with TAP 4
```
mnm678, line 172: I'm not 100% happy with requiring all repos to including the TAP 13 targets metadata, and am open to other suggestions that can allow for this interaction without too much added complexity.

tap13.md
```diff
@@ -148,16 +159,22 @@ use mode clearer in the specification.
...
+
+If a client is using TAP 4 to provide mapping metadata to multiple repositories,
+the TAP 13 targets mapping metadata will be applied to all repositories. This
+means that the targets metadata listed in the targets mapping metadata must
+be present on all repositories listed in TAP 4.
```
JustinCappos, line 177: I'm a bit confused here.  I think of the TAP 13 metadata as being per-repository inherently.  Are you proposing a single piece of metadata (from TAP 13) is applied to all repositories?

mnm678: Almost. This proposes that each repo needs to include metadata with the same name and keys. The metadata on each repo could include different delegations or packages.

I also considered adding a field to the TAP 4 metadata that lists which targets metadata to use, but this would require that all TAP 13 users implement all of TAP 4, even if they are only downloading from a single repository.

As stated above, I'm open to ideas that allow the client more control over which metadata is used on each repository without overly complicating the simple use case (a single repository user).

JustinCappos: Okay, why do you think the same name and keys for every repository is a good idea?  What is the use case?  (I'm sorry, but I think I'm missing something.)

From what I can think of, if you have multiple repos, you may want to pin keys for some (but not all of them) and also set a top level role for some (but not all of them), possible with different names

mnm678: The use case is that the client knows the packages/uploaders they want to trust on each repository, and so they create and upload targets metadata that includes this information to each repository. This targets metadata could delegate to other targets metadata file on the repo to include any combination of trusted roles, and could differ between repositories.

The downside would be if the client doesn't have upload permission to one or more repo that they are using, and so wouldn't be able to upload metadata with the same name.

What I'm not sure about is whether TAP 13 users will also be using TAP 4 often enough to justify a complication to this TAP (such as requiring TAP 4 mapping metadata with an additional "targets" field that points to the TAP 13 metadata) to allow for different TAP 13 metadata for each repo.

JustinCappos: I think it's also quite likely the person either won't have upload permission or won't be able to choose the same name on every repo.  For example, what if a company wants to have a pinned targets role on one repo that all of their developers use, but different developers want to be able to add other repos.  

I believe it would be much better to make TAP 13 metadata act on a per repo basis.

mnm678: I updated this section to add a section to TAP 4 metadata for a `targets-mapping`.

mnm678:
@JustinCappos @joshuagl Does the current text resolve the interaction with TAP 4?

tap13.md
```diff
@@ -148,16 +159,40 @@ use mode clearer in the specification.
...
+## Interaction with TAP 4
+
+If a client is using TAP 4 to provide mapping metadata to multiple repositories,
+they MAY provide a TAP 13 targets mapping for each repository.
+An optional `target-mapping` field will be added to TAP 4 to provide this mapping
+when TAP 13 is used. This field will contain:
+
+```
+"target-mapping": {
+   "repositories": [REPOSITORY_NAME, ...],
```
JustinCappos, line 181: If this is done per repository, what is the point of the repositories line?

mnm678: This lists the repositories that this target mapping applies to. In most cases there will only be one repository listed, with a new target mapping listed for each repo that needs one.

mnm678:
@JustinCappos @lukpueh @joshuagl Can I get a review on this?

tap13.md
```diff
@@ -148,16 +159,40 @@ use mode clearer in the specification.
...
+clients in the case of key loss or compromise for the top-level targets metadata file.  If it is
+operationally difficult to touch the clients, then the client may perhaps use a threshold of
+offline keys before delegating to a developer’s key.  [TAP 8](tap8.md) also provides support for
 cases where the key needs to be rotated or changed and the key is still accessible to the developer.
 
+## Interaction with TAP 4
+
+If a client is using TAP 4 to provide mapping metadata to multiple repositories,
+they MAY provide a TAP 13 targets mapping for each repository.
+An optional `target-mapping` field will be added to TAP 4 to provide this mapping
```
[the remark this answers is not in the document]
lukpueh: By the way, this is also true for the title of the TAP.

lukpueh (commented):
Thanks for the additions, @mnm678. It makes things clearer but also adds a couple of more questions (see inline).

tap13.md
```diff
@@ -148,16 +159,40 @@ use mode clearer in the specification.
...
-offline keys before delegating to a developer’s key.  [TAP 8](tap8.md) also provides support for 
+clients in the case of key loss or compromise for the top-level targets metadata file.  If it is
+operationally difficult to touch the clients, then the client may perhaps use a threshold of
+offline keys before delegating to a developer’s key.  [TAP 8](tap8.md) also provides support for
 cases where the key needs to be rotated or changed and the key is still accessible to the developer.
 
+## Interaction with TAP 4
+
+If a client is using TAP 4 to provide mapping metadata to multiple repositories,
+they MAY provide a TAP 13 targets mapping for each repository.
```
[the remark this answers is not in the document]
mnm678: This is based on the comments on this pr. The big issue is that when using multiple repositories you might want different TAP 13 metadata for each of these repos.

mnm678:
Thanks @lukpueh, sorry for the belated response

tap13.md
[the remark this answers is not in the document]
lukpueh: Just saw that there are a few more occurrences of `target` and `target's`, where I think we want to use `targets` and `targets'`. Would you mind going through the document and see if these need changes?

tap13.md
```diff
@@ -148,16 +159,42 @@ use mode clearer in the specification.
...
+operationally difficult to touch the clients, then the client may perhaps use a threshold of
+offline keys before delegating to a developer’s key.  [TAP 8](tap8.md) also provides support for
 cases where the key needs to be rotated or changed and the key is still accessible to the developer.
 
+## Interaction with TAP 4
+
+If a client is using TAP 4 to provide mapping metadata to multiple repositories,
+they MAY provide a TAP 13 targets mapping for each repository or group of repositories.
+An optional `targets_mappings` field will be added to TAP 4 to provide this mapping
+when TAP 13 is used. the mappings will be in resolved in order, so the first mapping
```
lukpueh, line 177:
```suggestion
when TAP 13 is used. The mappings will be resolved in order, so the first mapping
```
?

tap13.md
```diff
@@ -148,16 +159,42 @@ use mode clearer in the specification.
...
+will have higher priority than the second, and so on. This field will be resolved
+after the TAP 4 `mapping` field and will contain:
+
+```
+"targets_mappings": {
```
lukpueh, line 182: It still seems to me that `targets_mappings` should be a list and not a dictionary, right?

mnm678: I updated it to be a list of objects

tap13.md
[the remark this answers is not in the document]
mnm678: I fixed audited all the 'targets' and 'target'(s) in the document. They should all be correct now

lukpueh (approved):

joshuagl (approved):

joshuagl (approved):

PR #151, opened 2022-03-04
Add augmented reference implementation of TAP 12

mnm678:
[python-tuf 1.0](https://github.com/theupdateframework/python-tuf/releases/tag/v1.0.0) lets keyids be arbitrary strings as described in this TAP (discussed in theupdateframework/python-tuf#1084). As such, it can be used as an implementation of this TAP.

With this, are there any blockers to moving this TAP to accepted?

PR #152, opened 2022-03-04
Move TAP 12 to accepted

mnm678:
This pr proposes moving TAP 12 to accepted.

Requires #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,11 @@
 * [TAP 9: Mandatory metadata signing schemes](tap9.md)
 * [TAP 10: Remove native support for compressed metadata](tap10.md)
 * [TAP 11: Using POUFs for Interoperability](tap11.md)
+* [TAP 12: Improving keyid flexibility](tap12.md)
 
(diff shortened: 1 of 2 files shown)


joshuagl:
The Augmented Reference Implementation section of the TAP lists "TODO", how close are current python-tuf and go-tuf to the behaviour described here?

Once the TAP is Accepted, should we file issues (or pull requests) against those projects to match the TAP?

mnm678:
I believe the new python-tuf client is already compliant with this TAP as it does not verify keyids by the hash algorithm. That was added to the Augmented Reference Implementation section in #151.

There is an open [issue in go-tuf](https://github.com/theupdateframework/go-tuf/issues/232) for this TAP that we will want to address when this is accepted.

joshuagl:
Thank you. I _thought_ we had discussed it but couldn't see that reflected in the branch the PR was opened from? 🤷‍♂️

lukpueh (approved):

Issue #132, opened 2020-12-16
Discussion of TAP 15: Succinct hash bin delegations


joshuagl:
TAP 15: Succinct hash bin delegations is available at https://github.com/theupdateframework/taps/blob/master/tap15.md

Pull requests related to TAP 15: https://github.com/theupdateframework/taps/pull/120

Outstanding items for TAP 15:
* Can we make it more clear in the object format that a delegations object needs *either* `name` or `succinct_hash_delegations`, and `path_hash_prefixes` is optional. If the delegations object includes a `succinct_hash_delegations` field then any `path_hash_prefixes` field is ignored.
  * would be nice to do so in a way that suggests a relationship between `succinct_hash_delegations` and `path_hash_prefixes`
* Resolve on the use of `name` vs `bin_name_prefix` (probably through some experimenting on the PoC) see https://github.com/theupdateframework/taps/pull/120#discussion_r476656697

mnm678:
>     * Can we make it more clear in the object format that a delegations object needs _either_ `name` or `succinct_hash_delegations`, and `path_hash_prefixes` is optional. If the delegations object includes a `succinct_hash_delegations` field then any `path_hash_prefixes` field is ignored.
>       
>       * would be nice to do so in a way that suggests a relationship between `succinct_hash_delegations` and `path_hash_prefixes`

Looking back at this with fresh eyes, it might make more sense to include all three of these fields in the same definition to make the relationships more clear. A delegation may use either `succinct_hash_delegations` or (`name` and optional `path_hash_prefixes`). Which would look something like `"succinct_hash_delegations" : ... | "role_name" : {"name": ROLENAME, ("path_hash_prefixes": [ HEX_DIGEST, ...])}`.

Unfortunately, the above definition would be a backwards-incompatible change, so it may be better to leave it as is for now (as a change that does not affect non-succinct delegations), and set this as a goal for 2.0.0.

> 
>     * Resolve on the use of `name` vs `bin_name_prefix` (probably through some experimenting on the PoC) see [#120 (comment)](https://github.com/theupdateframework/taps/pull/120#discussion_r476656697)

This relates somewhat to the format of the delegations object. With the definition I proposed above, `bin_name_prefix` within the `succinct_hash_delegations` definition is more clear, but if we are leaving the current definition of the `name` field, we can discuss this further.

mnm678:
The discussion in theupdateframework/specification#156 may be relevant for finding a clear way to state the various options for required fields in this TAP.

[...]

PR #154, opened 2022-05-04
Re-design TAP 15 metadata format

mnm678:
Address some discussion from #132 by moving succinct hashed bin delegations to a new object within delegations.

In this version I opted to keep `path_hash_prefixes`, although I added a note that this field may be removed later backwards-incompatible way.

For simplicity, this change states that only one of `succinct_roles` or `roles` may be included. This seems to cover the most common use cases. Users with less common use cases could set up a complicated delegation tree if they would like to combine namespaced delegations and succinct hashed bin delegations, but it can't be done in the same delegation.

cc @jku @MVrachev @JustinCappos @trishankatdatadog @lukpueh

### DIFF
--- a/tap15.md
+++ b/tap15.md
@@ -128,7 +128,7 @@ that includes the following:
 "succinct_roles" : {
        "keyids" : [ KEYID, ... ] ,
        "threshold" : THRESHOLD,
-       "prefix_bit_length": BIT_LENGTH,
+       "bit_length": BIT_LENGTH,
        "name_prefix": NAME_PREFIX,
    }
 
@@ -192,7 +192,7 @@ With the addition of succinct hashed bins, the delegation will contain:
   "succinct_roles" : {
          "keyids" : [ KEYID, ... ] ,
          "threshold" : THRESHOLD,
-         "prefix_bit_length": BIT_LENGTH,
+         "bit_length": BIT_LENGTH,
          "name_prefix": NAME_PREFIX,
      },
 }
@@ -210,7 +210,7 @@ With the addition of succinct hashed bins, the delegation will contain:
     "succinct_roles" : {
         "keyids" : [ abc123 ] ,
         "threshold" : 1,
-        "prefix_bit_length": 16,
+        "bit_length": 16,
         "name_prefix" : "alice.hbd-",
     },
   }
--- a/tap15.md
+++ b/tap15.md
@@ -186,7 +186,7 @@ With the addition of succinct hashed bins, the delegation will contain:
       KEYID : KEY,
       ...
   },
-  "roles" : [
+  ("roles" : [
     {
       "name": ROLENAME,
       "keyids" : [ KEYID, ... ] ,
@@ -196,13 +196,13 @@ With the addition of succinct hashed bins, the delegation will contain:
       "terminating": TERMINATING,
     },
     ...
-  ],
-  "succinct_roles" : {
+  ],)
+  ("succinct_roles" : {
          "keyids" : [ KEYID, ... ] ,
          "threshold" : THRESHOLD,
          "bit_length": BIT_LENGTH,
          "name_prefix": NAME_PREFIX,
-     },
+     },)
 }
  ```
 
--- a/tap15.md
+++ b/tap15.md
@@ -142,7 +142,9 @@ previous delegations.
 
 When a delegation contains this field, it represents delegations to
 2^BIT_LENGTH bins that use the included keyids and threshold. All
-succinct hashed bin delegations will be non-terminating. The
+succinct hashed bin delegations will be non-terminating. If a user
+would like succinct delegations to be terminating, they may add the
+terminating flag in either the parent delegation or in the individual bins. The
 path_hash_prefixes and name for each
 bin will be determined using the BIT_LENGTH and NAME_PREFIX.
 
--- a/tap15.md
+++ b/tap15.md
@@ -161,7 +161,7 @@ The rolename of each bin will be determined by the bin number and the
 NAME_PREFIX listed in the `name_prefix` field of the delegation.
 The name will be structured as
 NAME_PREFIX-COUNT where COUNT is a hexadecimal value between 0 and
-2^BIT_LENGTH-1 (inclusive) that represents the bin number.
+2^BIT_LENGTH-1 (inclusive) that represents the bin number. This value will be zero-padded so that all rolenames will be of the same length.
 
 Only one of `succinct_roles` or `roles` may be specified in a
 delegation. If a role A would like to delegate to both `succinct_roles`
@@ -221,6 +221,8 @@ With the addition of succinct hashed bins, the delegation will contain:
     },
   }
  </code></pre>
+
+ The associated bins will be named `alice.hbd-0000`, `alice.hbd-0001`, ... `alice.hbd-FFFF`.
 
 # Security Analysis
 
--- a/tap15.md
+++ b/tap15.md
@@ -194,8 +194,8 @@ With the addition of succinct hashed bins, the delegation will contain:
       "terminating": TERMINATING,
     },
     ...
-  ],) |
-  ("succinct_roles" : {
+  ], |
+  "succinct_roles" : {
          "keyids" : [ KEYID, ... ] ,
          "threshold" : THRESHOLD,
          "bit_length": BIT_LENGTH,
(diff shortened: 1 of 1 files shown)


JustinCappos:
I think it is worth being explicit about how a maintainer would get the same functionality as supporting allowing succinct hash delegations within a single targets file.

tap15.md
```diff
@@ -121,21 +121,29 @@ calculated, see [this spreadsheet](https://docs.google.com/spreadsheets/d/10AKDs
...
 integer between 1 and 32 (inclusive).
 
+KEYID and THRESHOLD have the same definitions as in
+previous delegations.
+
 When a delegation contains this field, it represents delegations to
-2^BIT_LENGTH bins that use the keyid, threshold, path, and termination
-status from this delegation. The path_hash_prefixes and name for each
+2^BIT_LENGTH bins that use the included keyids and threshold. All
+succinct hashed bin delegations will be terminating. The
```
jku, line 145: Hmm, I wonder if this is correct. It seems to prevent the idea of using e.g. multiple succinct delegations (say, "targets" delegating to "succinct1" and "succinct2" using normal non terminating delegations, and both of those contain a succinct_roles)? I think succinct delegations could even be always non-terminating as the parent delegation could specify terminating if it wants to...

(I admit I still don't know if multiple succinct delegations has a use case, just pointing this out)

jku (commented):
this looks about right to me. Martin is working on a prototype I believe so we should get some implementation comments soon.

tap15.md
mnm678: Hmm, I think we can safely remove the option for multiple succinct delegations using the same workaround as for using both `roles` and `succinct_roles` (setting up multiple intermediate roles). I don't think that this will be common, and having only one succinct delegation per role simplifies the mechanism a lot.

That said, I don't have a strong preference for making these always terminating, it just seemed like a way to simplify the metadata.

jku: I wasn't trying to suggest multiple succinct delegations in one metadata; I was talking about a "normal" role that delegates to two roles that are both succinct. If they are automatically terminating, then that setup makes no sense: I think they should either always be non-terminating (as the parent delegation can always terminate if it wants to) or should include the terminating field.

MVrachev:
> this looks about right to me. Martin is working on a prototype I believe so we should get some implementation comments soon.

I have implemented the prototype and left a comment about it here: https://github.com/theupdateframework/python-tuf/issues/1943#issuecomment-1121282526.

mnm678:
@MVrachev @jku @lukpueh 

I updated the TAP per our discussion today to change `prefix_bit_length` to `bit_length` and clarify a few things.

tap15.md
```diff
@@ -168,39 +181,45 @@ files should be uploaded before the metadata that delegates to them.
...
+    "succinct_roles" : {
         "keyids" : [ abc123 ] ,
-        "succinct_hash_delegations" : {
-          "delegation_hash_prefix_len" : 16,
-          "bin_name_prefix" : alice.hbd-
-        },
         "threshold" : 1,
-        "terminating": false,
+        "bit_length": 16,
+        "name_prefix" : "alice.hbd-",
```
lukpueh, line 222: 💯  for succinct names!

MVrachev (changes requested):
There was a discussion about TAP 15 with TAP author @mnm678, @jku, @lukpueh, @abs007 and me where it was decided the following:
1. Succinct hash delegation and `roles` would be separated. `roles` will only be responsible for standard roles and no succinct hash delegations.
2. As a result of the first point `keyids` and `threshold` attributes were added to succinct hash delegation.
3. Succinct hash delegations and standard roles (the `roles` attribute) should not exist together in one delegations object.
4. Only one succinct hash delegation must exist. 
5. There are a couple of renames that we agreed on:
- `succinct_hash_delegations` → `succinct_roles`
- `delegation_hash_prefix_len` -> `bit_length`
- `bin_name_prefix` → `name_prefix`

tap15.md
```diff
@@ -121,21 +121,31 @@ calculated, see [this spreadsheet](https://docs.google.com/spreadsheets/d/10AKDs
...
+
 When a delegation contains this field, it represents delegations to
-2^BIT_LENGTH bins that use the keyid, threshold, path, and termination
-status from this delegation. The path_hash_prefixes and name for each
+2^BIT_LENGTH bins that use the included keyids and threshold. All
+succinct hashed bin delegations will be non-terminating. If a user
+would like succinct delegations to be terminating, they may add the
+terminating flag in either the parent delegation or in the individual bins. The
+path_hash_prefixes and name for each
 bin will be determined using the BIT_LENGTH and NAME_PREFIX.
```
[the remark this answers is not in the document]
mnm678: Correct. I was referring to how the rolename is generated, but I think it's just confusing to put this here when it's already described a couple of paragraphs below

tap15.md
```diff
@@ -121,22 +121,30 @@ calculated, see [this spreadsheet](https://docs.google.com/spreadsheets/d/10AKDs
...
 integer between 1 and 32 (inclusive).
 
+KEYID and THRESHOLD have the same definitions as in
+previous delegations.
+
 When a delegation contains this field, it represents delegations to
-2^BIT_LENGTH bins that use the keyid, threshold, path, and termination
-status from this delegation. The path_hash_prefixes and name for each
-bin will be determined using the BIT_LENGTH and NAME_PREFIX.
+2^BIT_LENGTH bins that use the included keyids and threshold. All
```
joshuagl, line 144: Nit: worth emphasising that all of the 2^BIT_LENGTH bins use the same keyids and threshold?
```suggestion
2^BIT_LENGTH bins which all use the specified keyids and threshold. All
```

tap15.md
```diff
@@ -168,39 +179,45 @@ files should be uploaded before the metadata that delegates to them.
...
+    {
+      "name": ROLENAME,
+      "keyids" : [ KEYID, ... ] ,
+      "threshold" : THRESHOLD,
+      ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+      "paths" : [ PATHPATTERN, ... ]),
+      "terminating": TERMINATING,
+    },
+    ...
+  ],)
```
joshuagl, line 197:
```suggestion
  ],) |
```
as it's one of `roles` or `succinct_roles`

joshuagl (dismissed):
Big thanks to everyone involved in this effort. It's awesome to see the TAP evolve based on an effort to implement.

Succinct hashed bin delegations now look much easier to implement and much easier to recognise in metadata. 🎉

joshuagl (approved):

tap15.md
```diff
@@ -168,39 +179,45 @@ files should be uploaded before the metadata that delegates to them.
...
+      "name": ROLENAME,
+      "keyids" : [ KEYID, ... ] ,
+      "threshold" : THRESHOLD,
+      ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+      "paths" : [ PATHPATTERN, ... ]),
+      "terminating": TERMINATING,
+    },
+    ...
+  ],) |
+  ("succinct_roles" : {
```
lukpueh, line 198:
```suggestion
  ], |
  "succinct_roles" : {
```

Not sure if this is correct, but it's consistent with `("path_hash_prefixes": ... | "paths": ...)` above, where we don't group the individual OR'ed elements but the whole thing.

tap15.md
```diff
@@ -168,39 +179,45 @@ files should be uploaded before the metadata that delegates to them.
...
-    "roles" : [{
+    "succinct_roles" : {
         "keyids" : [ abc123 ] ,
-        "succinct_hash_delegations" : {
-          "delegation_hash_prefix_len" : 16,
-          "bin_name_prefix" : alice.hbd-
-        },
         "threshold" : 1,
-        "terminating": false,
+        "bit_length": 16,
```
lukpueh, line 219: The TAP does not mandate this, but for the implementation we decided that it makes sense to have fixed length role names within one succinct delegation, by zero-padding the bin number part of the role name, where necessary.

If there is an implicit recommendation in the TAP (you do apply zero padding in the example above; maybe there  should be an explicit recommendation?), then `bit_length: 16` would entail 4 digit bin numbers, i.e. `alice.hbd-0000` to `alice.hbd-ffff`.

lukpueh (dismissed):
LGTM modulo one minor formatting/consistency nit and a metadata example issue that could be addressed in a follow-up PR.

lukpueh (approved):

joshuagl (approved):

PR #155, opened 2022-06-09
TAP 15: misc improvements and fixes

lukpueh:
This PR contains a few random improvements and fixes of things that I noticed in my last reading of the document. The commit log has details about the changes. When reviewing the diff I recommend disabling whitespace changes.

### DIFF
--- a/tap15.md
+++ b/tap15.md
@@ -158,8 +158,8 @@ values starting with 000, the second bin would include binary values starting
 with 001, the third 010, then 011, 100, 101, 110, 111.
 
 The rolename of each bin will be determined by the bin number and the
-NAME_PREFIX listed in the `name_prefix` field of the delegation.
-The name will be structured as
+NAME_PREFIX listed in the `name_prefix` field of the delegation,
+separated by hyphen (`-`). The name will be structured as
 NAME_PREFIX-COUNT where COUNT is a hexadecimal value between 0 and
 2^BIT_LENGTH-1 (inclusive) that represents the bin number. This value will be zero-padded so that all rolenames will be of the same length.
 
@@ -217,7 +217,7 @@ With the addition of succinct hashed bins, the delegation will contain:
         "keyids" : [ abc123 ] ,
         "threshold" : 1,
         "bit_length": 16,
-        "name_prefix" : "alice.hbd-",
+        "name_prefix" : "alice.hbd",
     },
   }
  </code></pre>
--- a/tap15.md
+++ b/tap15.md
@@ -45,29 +45,37 @@ follows a common pattern.
 This TAP supports the following use case:
 
 Suppose a single targets metadata file contains a very large number of
-delegations or target files. The owner of this targets metadata file wishes to
-reduce the metadata overhead for clients, and so uses hashed bin
-delegations. They will use the same key to sign each bin delegation.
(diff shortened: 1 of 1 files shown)


tap15.md
```diff
@@ -45,92 +45,91 @@ follows a common pattern.
...
+
+For a repository with 50,000,000 target files using the existing hashed
+bin delegation technique with a number of bins of 2,048 and 32-bit
+keyids the snapshot and targets metadata overhead would be around
 1,600,000 bytes for each target. Using succinct hashed bin delegations,
 the snapshot and targets metadata overhead for a target can be reduced
 to about 550,000 bytes. For more detail about how these overheads were
-calculated, see [this spreadsheet](https://docs.google.com/spreadsheets/d/10AKDsHsM2mmh45CWCNFxihJ9f-SP6gXYv7WcWpt-fDQ/edit#gid=0).
+calculated, see [this
+spreadsheet](https://docs.google.com/spreadsheets/d/10AKDsHsM2mmh45CWCNFxihJ9f-SP6gXYv7WcWpt-fDQ/edit#gid=0).
```
lukpueh, line 119: @mnm678, it seems like you are comparing metadata overheads with different numbers of bins for traditional hbd and succinct hbd. Is this intended?

mnm678: Yes, I tried to fairly optimize both strategies to minimize the metadata overhead, which lead to different numbers of bins.

mnm678 (approved):
thanks @lukpueh, this looks great

trishankatdatadog:
Need anyone else to review here?

tap15.md
lukpueh: Interesting, I always thought that the number of targets alone determines the optimal number of bins, but it looks like in traditional hbd the size of the delegating metadata undermines the positive effect of many bins.

lukpueh: Btw. I saw that in your calculations you use 32 bytes for keyids, not 32 bits as mentioned in the text. And shouldn't it actually be 64 bytes, since we usually use the hexdigest of sha256?

lukpueh: I should probably take out the bin number above in the overhead calculations, given that we use two different ones. Or should we point out that they have different optimal numbers of bins? 

Maybe it's not so important. I suppose the key here is that the size of the delegating metadata grows linearly with the number of bins for traditional hbd, whereas it does not grow for succinct hbd.

joshuagl (approved):
Nice improvements, thanks @lukpueh

tap15.md
lukpueh: Any opinions about how to proceed with this comparison paragraph? The options I see:
1. add detail about (mention that two different bin numbers where used and why)
2. remove detail (don't mention bin numbers)
3. reword (e.g. don't compare full metadata overhead, but delegating metadata sizes only and point out "linear vs. no growth")

I lean towards 3 and am happy to update this PR.

cc @trishankatdatadog, @mnm678, @joshuagl

mnm678: I agree. I think 3 clarifies without confusing the reader

lukpueh: Okay, I'll submit another PR for this shortly. Let's merge what we have, given that we already have two approvals.

PR #156, opened 2022-06-20
Add TAP supporting content addressed targets

adityasaky:
This TAP proposes supporting Merkle DAG objects / nodes as targets in TUF metadata. This allows us to extend TUF's protections to various popular applications like Git and other content addressable systems like IPFS and OSTree.
- One key change proposed in this TAP is that the ecosystem or application decides how to calculate the hash of its artifact, and this value is re-used.
- Generally speaking, to record the hash of a non-file artifact, we need some representation of it. This TAP indirectly proposes using the representation used by the application or ecosystem in question by re-using the hash based identifier.

The TAP goes into some detail about the properties the application of ecosystem under consideration must possess to ensure the integrity of the hash values. Feedback as a whole is welcome! There are some discussion points in line that I'm also going to copy here:
- Should conforming implementations also implement base TUF for regular files?
- What makes sense for `length` for Git objects? The commit object files pre-compression?
    - This is a use-case specific question that's perhaps better handled when considering the ecosystem specifically rather than this TAP as a whole, but I also wonder if we can provide more guidance on fields that are affected by these changes.

### DIFF
--- a/POUFs/TAF-POUF/pouf2.md
+++ b/POUFs/TAF-POUF/pouf2.md
@@ -1,0 +1,140 @@
+* POUF:
+* Title: The Archive Framework
+* Version: 1
+* Last-Modified:
+* Author: Renata Vaderna
+* Status: Draft
+* TUF Version Implemented:
+* Implementation Version(s) Covered:
+* Content-Type: text/markdown
+* Created:
+
+# Abstract
+
+This POUF describes the protocol, operations, usage, and formats for the
+implementation of TUF designed to distribute Git repositories. This instance is
+known as The Archive Framework or TAF and leverages TAP-19 that adds support in
+TUF for content addressable artifacts and their native hashing routines.
+
+# Protocol
+
+This POUF currently uses a subset of the JSON object format, with floating-point
+numbers omitted. When calculating the digest of an object, we use the
+["canonical JSON" subdialect](http://wiki.laptop.org/go/Canonical_JSON) and
+implemented in securesystemslib. As TAF uses the TUF reference implementation
+for metadata generation, it implicitly depends on the reference implementation
+POUF’s protocols.
+
+Metadata and target files are stored in a git repository referred to as an
+authentication repository. An authentication repository contains information
+needed to securely clone and update other git repositories (referred to as
+target repositories), including their URLs and additional custom data that can
+be used by TAF's implementers. This is specified in special target files,
+`repositories.json` and `mirrors.json` - regular TUF  target files (whose
+lengths and hashes are stored in `targets.json` and which are signed by the
+top-level targets role) which are of special importance to TAF.
+
+TAF focuses on protecting git repositories from unauthorized pushes. It is
+designed to record valid commits for target repositories in the authentication
+repository. If an attacker manages to compromise a user who has write access to
+protected repositories and creates new commits, TAF will detect this by
+comparing these new commits to a list of valid commits specified in the
+authentication repository. In order to register these new commits, the attacker
+has to modify TUF metadata files as well. Unless they also gain access to
+`targets` or `root` keys, they cannot do this without the attempt being
+detected by TAF. TAF's users are encouraged to store signing keys for the
+`target` and `root` roles offline on hardware tokens to ensure their safety.
+
+In essence, while TAF ensures the validity of commits in target repositories,
+it makes no claims about the integrity of their contents. TAF relies on Git's
+default artifact integrity protection. Git, however, still primarily relies on
+SHA-1, even though it has been proven that this hash function is vulnerable to
+collision attacks. If an attacker manages to gain access to the original target
+repositories, they could potentially exploit the SHA-1 weakness. On the other hand,
+authentication repositories store lists of valid URLs of target repositories,
+ensuring that users are protected from _mirrors_ of legitimate repositories presenting
+a colliding artifact in place of the original artifact. These URLs are defined manually
+and can only be modified by someone who has a `targets` key. If a target repository
+is not owned by the same person or organization that is setting up an authentication
+repository, it is a hard requirement to directly contact the owner who will be
+able to confirm authenticity of the repository in question.
+
+In order to take advantage of TAF's validations, a client can download an
+authentication repository and all of the referenced repositories by running
+TAF’s updater and specifying the authentication repository’s URL. Repositories
+are cloned and updated using git. The updater runs validation before
+permanently storing anything on the client’s machine:
+*   An authentication repository is cloned as a bare repository inside the
+    user’s temp directory (so no worktree is checked out)
+*   This repository is then validated. Metadata and target files are read using
+    `git show`
+*   If validation is successful, the repository is cloned/new changes are
+    pulled, once again using Git
+
+
+# Operations
+
+WIP
+
+# Usage
+
+In order to use the system, it is necessary to set up an authentication
+repository - initialize a TUF repository (generate TUF metadata files and sign
+them using offline keys, either loaded from the filesystem or YubiKeys) and
+commit the changes. TAF contains a command line interface which can be used to
+create and update authentication repositories (add new target files and signing
+keys, extend expiration dates of metadata files, generate keystore files and
+set up YubiKeys). For example, a new authentication repository can be created
+using the `taf repo create repo_path` command. Detailed explanation and
+instructions are available in the official documentation
+https://github.com/openlawlibrary/taf/blob/master/docs/quickstart.md.
+
+TAF's main purpose is to provide archival authentication, which means that it
+is not only the current state of the repositories that is validated -
+correctness of all past versions needs to be checked as well. A state is valid
+if the authentication repository at a certain revision (commit) is a valid TUF
+repository and target repositories are valid according to the data defined in
+the authentication repository. So, if `targets.json` is updated,
+`snapshot.json` and `timestamp.json` need to be updated in the same commit or
+the repository will not be valid. This ensures that a client cannot check out
+an invalid version of the repository. Moreover, validation of the
+authentication repository also ensures that versions of metadata files in older
+revisions are lower than in newer revisions.  Once the metadata and target
+files are validated, they are used to check correctness of the referenced git
+repositories - do actual commits in those repositories correspond to the
+commits listed in the authentication repository.
+
+To sum up, TAF extends the reference implementation by storing the metadata and
+target files in a git repository and making sure that all changes are committed
+after every valid update.
+
+# Formats
+
+The metadata generated to support Git repositories are largely the same as
+those described in the TUF specification and POUF-1. The key difference is in
+the enumeration of Targets.
+
+Firstly, while TUF identifies individual targets by their location relative to
+the mirror’s base URL, this POUF uses a URI to identify the specific Git
+namespace in a target repository. This URI is also used to locate the
+repository itself. The URI must use `git` as the scheme, clearly indicating
+that the entry pertains to a Git object.
+
+For metadata of some repository, a Targets entry is expected to map to a
+specific Git branch or tag. For each entry, whether a branch or a tag, a hash
+value must be recorded that clearly identifies the expected commit at the
+corresponding Git reference. As per TAP-19 that adds support for content
+addressable systems and their native hashing routines, instead of calculating
+the hash afresh for a particular Git reference, the identifier of the commit at
+the tip of the branch or that the tag points to must be used.
+
+Apart from the hashes and custom field, a Targets entry is also expected to
+record the length of the artifact. This length is vital in avoiding endless
+data attacks. However, there is no clear mapping of a length for a particular
+commit object. Therefore, for Git specific implementations, the length field
+may be omitted. However, implementations are free to choose sane limits for how
+much data is fetched when pulling from a Git repository.
+
+# Security Audit
+
+None yet.
--- a/tap19.md
+++ b/tap19.md
@@ -1,0 +1,472 @@
+* TAP: 19
+* Title: Content Addressable Systems and TUF
+* Version: 1
+* Last-Modified: 25/04/2023
+* Author: Aditya Sirish A Yelgundhalli, Renata Vaderna, John Ericson
+* Type: Standardization
+* Status: Draft
+* Content-Type: markdown
+* Created: 09/05/2022
+* Requires: TAP-11
+* +TUF-Version:
+* +Post-History:
+
+# Abstract
+
+This TAP explores how TUF can be adapted to content addressed ecosystems that
+have in-built integrity checks for their artifacts. While the TUF specification
+supports verifying artifact integrity, it also describes many other semantics
+such as key distribution and the ability to delegate trust to different
+entities. Essentially, content addressed systems such as Git, IPFS, and OSTree
+have artifact integrity capabilities which can be complemented by all of TUF's
+other features.
+
+# Motivation
+
+The current TUF specification requires `targets` to be files--TUF is designed
+around the assumption that the artifacts being distributed are "regular"
+files. However, emerging applications and ecosystems that defy this assumption
+can still greatly benefit from TUF's security properties.
+
+Content addressed systems are those in which objects are addressed or identified
+by a function of their contents. Typically, each artifact is addressed using a
+cryptographic hash of its contents. Due to this characteristic, these systems
+typically enforce the verification of artifact integrity intrinsically. Some
+examples of content addressed systems are Git, the Interplanetary File System
+(IPFS), and OSTree. In Git, if some object with a particular content address is
+overwritten in the Git object store, any operations that _use_ the corrupted
+object fail because of a hash mismatch. IPFS provides interfaces to store
+artifacts at which point their hashes are calculated. These hashes can then be
+used to fetch the corresponding artifacts from IPFS. Similar to Git, a
+corruption of an object in the store results in failures when attempting to use
+that artifact.
+
+## Use Case 1: Open Law Library's The Archive Framework
+
+The Open Law Library is an open access publisher that makes laws freely
+accessible to governments and their citizens. They build tools that help
+governments with the drafting, codifying, and publishing aspects of the
+legislative process.
+
+The organization have developed a variant of TUF called The Archive Framework
+(TAF), designed to support Git repositories as targets rather than regular
+files. TAF uses a stand-in file representing each repository which records the
+specific commit ID. This file is then used as a target in TUF metadata. This is
+redundant as the file is only used to determine the correct Git object to
+use--TAF already relies on Git's default content integrity protections. By
+moving this information into the TUF metadata, this redundancy can be
+eliminated.
+
+This particular use case can be generalized to supporting any Git repositories
+as Targets.
+
+## Use Case 2: IPFS Artifacts
+
+IPFS provides a peer-to-peer protocol to store and transfer files. Every file is
+identified by a hash. Note that the identifier is not the hash of the file
+itself. IPFS represents each file as a series of individually addressed blocks
+and the identifier of the file as a whole encompasses all of these blocks.
+
+By supporting IPFS as a way to distribute artifacts from a repository, all of
+TUF's security properties can be applied except for the artifact integrity
+itself. When fetching a target, the client would verify all of the TUF metadata
+and then use the targets entry to determine which IPFS artifact to fetch.
+
+Finally, IPFS can be used to store the TUF metadata itself. TUF's Timestamp role
+can, in that case, address the latest Snapshot role using its IPFS address,
+Snapshot roles can similarly address Targets roles using their corresponding
+IPFS addresses, and so on.
+
+## Use Case 3: Distributing Artifacts Using OSTree
+
+OSTree, or libostree, is a library and tool that applies a Git-like model for
+entire bootable filesystem trees. The project includes utilities for deploying
+these images. It follows a similar principle as Git, using hash values to build
+content addressability. OSTree is used by a variety of projects such as Flatpak,
+Fedora CoreOS, and Automotive Grade Linux.
+
+Some of these OSTree adopters use it to generate and deliver software updates.
+For example, Toradex, a manufacturer of IoT boards, uses OSTree to deliver
+updates of their operating system TorizonCore to their devices. Toradex has also
+adopted the Uptane standard which implements TUF for automotive and other
+embedded device use cases. In essence, Toradex's deployment implements the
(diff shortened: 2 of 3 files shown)


trishankatdatadog:
Cc @erickt

adityasaky:
Hi Zack, thanks for your comments! I want to note some thoughts:

* I agree that Merkle DAG objects are a subset of entities we can record in TUF metadata. Initially, the idea with this specific
 TAP was to not clear the way for all non-file entities (that would require something like in-toto's ITE-4: https://github.com/in-toto/ITE/blob/master/ITE/4/README.adoc). In most cases, we'd have to specifically define how to calculate the hashes of these abstract entities, there may well be nothing we can use as is. ITE-4 handles that using something called a "hashable representation".
* The reason it talks about Merkle DAG ecosystems specifically is because apart from opening the door to non-file entities, it also allows for hash values that the TUF implementation didn't explicitly calculate, but was instead _provided with_. While there are varying degrees of verifiability for these values (it's quite straightforward to verify a Git commit's ID for example), one other thing this draft says is that the valid existence of a node in the graph can be used for that node's verification (Git does this well, for example, erroring out when there are invalid / tampered commit objects). This is where the idea of trusted ecosystems that validate their Merkle DAGs came from, given regular auditing of these applications / ecosystems to ensure new changes haven't undermined the assumptions made.
* We may not want to go this route for any ecosystems, instead requiring values to be explicitly validated each time. In this case, the TUF implementation would likely have to be aware of the nuances of the ecosystem in greater detail than it'd need to verify the existence of a node in a trusted application. I think this could get quite complicated when we get to "storage backends" like IPFS. I'm currently working on a python-tuf proof-of-concept that's looked at Git so far, but IPFS is what I want to play with next.

There was interest in the last community meeting for something like ITE-4, i.e., more open to non-file entities than what is here, so you're not alone in that regard.

merkle-dag-targets.md
```diff
@@ -0,0 +1,474 @@
...
+locate the repository itself. So, when the Merkle DAG in question is that of a
+Git repository, the target identifier should point to the root of the
+repository as a whole.
+
+```
+git:<repo identifier>
+```
+
+With Git in particular, the URI structure can be extended to contain more
+information. For example, the URI could record a specific _branch_ or _tag_.
```
erickt, line 153: Could you define the semantics of listing a branch of a tag, since those are not content addressed? Or state that this is a non-normative example of a possible extension?

merkle-dag-targets.md
```diff
@@ -0,0 +1,474 @@
...
+
+For example, every Git repository contains a Merkle DAG, in which every node is
+a commit object, and each commit has a unique identifier generated using SHA-1.
+However, while this identifier can point to the commit, it is insufficient to
+locate the repository itself. So, when the Merkle DAG in question is that of a
+Git repository, the target identifier should point to the root of the
+repository as a whole.
+
+```
+git:<repo identifier>
```
erickt, line 149: Could we further flesh out the repo identifier? It sounds like the identifier could be just a unique name, and it would be fine for the client to use a local private mirror. 

Or, are you using git and ipfs as non-normative examples, where this doc proposes how TUF can work with systems like git and ipfs, but the actual integration would be formally designed later using a pouf-like process?

merkle-dag-targets.md
```diff
@@ -0,0 +1,474 @@
...
+   "length" : LENGTH,
+   "hashes" : {ALG: HASH, ...},
+   ("custom" : CUSTOM)
+}
+```
+
+While this format does not require any significant changes, the techniques used
+to populate their values for non-file targets must be carefully designed.
+
+The opaque `custom` field requires no change to make the changes in this TAP
```
erickt, line 190: This is what we do with fuchsia. The target path is the name of the package, and we have a “merkle” entry in the custom field. Our implementation doesn’t actually try to download the file using the normal TUF methods, we are just query the targets metadata for the content addressed merkle, then download the rest of the package blobs from our blob repository.

merkle-dag-targets.md
```diff
@@ -0,0 +1,474 @@
...
+possible.
+
+Discuss: what makes sense for `length` in git repos? The commit object files
```
erickt, line 193: If there’s no logical “length” for git, rather than using “custom”, it might be worthwhile having a target “type” field, where a “git” entry might only have the repo and object hash, and “fuchsia” could have the merkle and length. The default could be an optional “file” scheme using the current definition.

merkle-dag-targets.md
adityasaky: It is more of the latter certainly, I'll make a chance to that effect. I believe that should also address https://github.com/theupdateframework/taps/pull/156#discussion_r928272048.

adityasaky: That said, for a branch or tag entry, the hash recorded is still content addressed to a specific commit. It's essentially discussing that these semantics can be used to record the commit of the tree trusted for a branch or a tag. However, as this is better elaborated formally in a POUF, should I take the bits about branches and tags out entirely?

merkle-dag-targets.md
adityasaky: While using Git as an example, shall I update it to say that not all fields may be necessary and it must be judged on a case-by-case basis? If I'm understanding it correctly, we have no backwards compatibility issues if the TAP also says that for any implementation also supporting regular files, the fields in the specification must be followed, right?

cc @mnm678

merkle-dag-targets.md
adityasaky: If I'm understanding this correctly, the `merkle` entry in `custom` contains the node's hash, yes? Do you leave the `hashes` object empty?

POUFs/TAF-POUF/pouf2.md
```diff
@@ -0,0 +1,53 @@
...
+
+# Protocol
+
+This POUF currently uses a subset of the JSON object format, with floating-point numbers omitted. When calculating the digest of an object, we use the ["canonical JSON" subdialect](http://wiki.laptop.org/go/Canonical_JSON) and implemented in securesystemslib. As TAF uses the TUF reference implementation for metadata generation, it implicitly depends on the reference implementation POUF’s protocols.
+
+Metadata and target files are stored in a git repository referred to as an authentication repository.  An authentication repository contains information needed to securely clone and update other git repositories (referred to as target repositories), including their URLs and additional custom data that can be used by TAF's implementers. This is specified in special target files, `repositories.json` and `mirrors.json` - regular TUF  target files (whose lengths and hashes are stored in `targets.json` and are singed by the top-level targets role) which are of special importance to TAF.
```
lukpueh, line 20: typo
```suggestion
Metadata and target files are stored in a git repository referred to as an authentication repository.  An authentication repository contains information needed to securely clone and update other git repositories (referred to as target repositories), including their URLs and additional custom data that can be used by TAF's implementers. This is specified in special target files, `repositories.json` and `mirrors.json` - regular TUF  target files (whose lengths and hashes are stored in `targets.json` and are signed by the top-level targets role) which are of special importance to TAF.
```

lukpueh:
This is interesting. Which of TUFs security properties do we actually care for here? Is TUFs integrity protection even relevant, when targets are content addressable? In other words, does the client need to verify the target hash at all, or is it enough that the target path is in targets metadata?

adityasaky:
> Which of TUFs security properties do we actually care for here?

That's a great point and I think emphasizing that in the text will greatly help. With content addressed systems, we care about all of TUF's properties minus artifact integrity. Let me take a pass on that.

Also note that we've proposed a prototype of this TAP as a GSoC 2023 task.

adityasaky:
@lukpueh I reworked the TAP to focus on TUF properties that matter outside of artifact integrity validation. LMK what you think!

tap19.md
```diff
@@ -0,0 +1,360 @@
...
+(TAF), designed to support Git repositories as targets rather than regular
+files. TAF uses a stand-in file representing each repository which records the
+specific commit ID. This file is then used as a target in TUF metadata.
+
+This particular use case can be generalized to supporting any Git repositories
+as Targets.
+
+## Use Case 2: IPFS as a Backend for Targets
+
+TODO: John?
```
mnm678, line 60: It'd be nice to get this added or removed before merging as a draft.

adityasaky: cc @ericson2314

trishankatdatadog:
(Sorry, just wanted to say pls count me out of reviewing this TAP now as I will be on a few weeks of PTO. Thanks!)

tap19.md
```diff
@@ -0,0 +1,360 @@
...
+single identifier--that of the root of the tree.
+
+# Specification
+
+The TUF specification uses file hashes in a number of contexts. All Targets
+metadata entries are expected to record hashes of the corresponding entries
+using one or more algorithms. Snapshot metadata records the hashes of all
+Targets metadata considered valid at the time of its issuance, and Timestamp
+metadata points to the currently valid Snapshot metadata file. In each of these
+contexts, TUF operates with the assumption that the artifacts whose hashes are
```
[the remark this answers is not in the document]
adityasaky: We'll update this when we tackle IPFS specifically with how we can use it to store TUF metadata as well.

tap19.md
```diff
@@ -0,0 +1,360 @@
...
+
+This TAP explores how TUF can be adapted to content addressed ecosystems that
+have in-built integrity checks for their artifacts. While the TUF specification
+supports verifying artifact integrity, it also describes many other semantics
+such as key distribution and the ability to delegate trust to different
+entities. Essentially, content addressed systems such as Git, IPFS, and OSTree
+have artifact integrity capabilities which can be complemented by all of TUF's
+other features.
+
+# Motivation
```
jkjell, line 24: Is leveraging an OCI registry a valid use case for this TAP?

tap19.md
```diff
@@ -0,0 +1,360 @@
...
+## Identifying the Target
+
+As TUF is centered around regular file artifacts, each entry uses a path that is
+relative to the repository's base URL. In content addressed systems, the name is
+not as straightforward, and can instead be ecosystem specific. For example, in
+the Git use case, the entry's name can identify the repository and the Git ref
+the entry applies to.
+
+Therefore, the entry must clearly identify the ecosystem it pertains to. This
+TAP proposes using RFC 3986's URI structure for the entry's identifier.
```
jkjell, line 110: RFC 3986 states 

> This specification does not require that a URI persists in identifying the same resource over time

However, it does not prevent the contrary.  Is that worth calling out?  The `hier-part` examples below do not represent a consistent over-time URI.

jkjell (commented):
This looks great and sounds really interesting!  I dropped a bunch of noob questions in my review.  One of the things I was quite sure where to put was around the common and implicit nature of many of these content addressable systems to be used in a distributed environment.  This often leads to separate integrity checks at the "server" and the "client" of the application.  I don't know if that needs to be more explicitly addressed or would just be simply covered in the "Security Assessment" of the ecosystem.

POUFs/TAF-POUF/pouf2.md
```diff
@@ -0,0 +1,116 @@
...
+Firstly, while TUF identifies individual targets by their location relative to
+the mirror’s base URL, this POUF uses a URI to identify the specific Git
+namespace in a target repository. This URI is also used to locate the repository
+itself. The URI must use `git` as the scheme, clearly indicating that the entry
+pertains to a Git object.
+
+For metadata of some repository, a Targets entry is expected to map to a
+specific Git branch or tag. For each entry, whether a branch or a tag, a hash
+value must be recorded that clearly identifies the expected commit at the
+corresponding Git reference. As per TAP-X that adds support for content
```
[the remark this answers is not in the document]
adityasaky: Yep, I recently numbered this one. I'll edit this file, good catch!

tap19.md
adityasaky: Yes, I think depending on the ecosystem in question, the URI may or may not always point to the same resource. The TAP should call this out and defer to the POUF to set expectations.

tap19.md
```diff
@@ -0,0 +1,360 @@
...
+
+The opaque `custom` field requires no explicit changes. An ecosystem may choose
+to define some specific fields within it, and this must be communicated in the
+corresponding POUF.
+
+The `length` field is an integer that captures the length in bytes of the
+target. This field may or may not be relevant, depending on the ecosystem. The
+POUF must specify how `length` is to be parsed.
+
+Similarly, the `hashes` field me be unnecessary if the target identifier
```
lukpueh, line 153: typo?

tap19.md
```diff
@@ -0,0 +1,360 @@
...
+entire, bootable filesystem trees. The project includes utilities for deploying
+these images. It follows a similar principle as Git, using hash values to build
+content addressability. OSTree is used by a variety of projects such as Flatpak
+and Fedora CoreOS.
+
+One key use of OSTree is for packaging and distribution operations. With
+OSTree, a package manager can distribute an entire filesystem tree. In some
+cases, such a tree can be the artifact itself, say an operating system image,
+but OSTree also makes it possible to distribute multiple artifacts using a
+single identifier--that of the root of the tree.
```
lukpueh, line 74: nit: I think the motivation section could still be stronger

Right now it says that there are applications that could greatly benefit from TUF, but it does not say how. Instead it goes into detail about an important TUF property that these applications don't need. Similarly, it describes two exemplary use cases, but not how they would benefit from TUF. On the contrary, use case 1 seems to work even without this TAP. Why is the existing workaround not viable?


fom TAP 1:
> The motivation is critical for TAPs that want to change TUF. It should clearly explain why the existing framework specification is inadequate to address the problem that the TAP solves.

lukpueh (commented):
Thanks for the updates, @adityasaky, this looks a lot better!

tap19.md
adityasaky: I _believe_ so from when I last looked through the specifics but I want to review it again before we include it. Does it read to you as the right fit for OCI registries?

tap19.md
adityasaky: Updated the text, thanks for highlighting this!

tap19.md
adityasaky: Update: I think this applies to artifacts stored in an OCI registry, yes.

tap19.md
adityasaky: Updated.

tap19.md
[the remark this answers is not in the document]
adityasaky: Updated.

tap19.md
adityasaky: Done!

renatav (commented):
I think that there are a couple of things that should be covered by the POUF according to the TAP that we haven't addressed in the TAF POUF, like backwards compatibility. Is that a blocker?

tap19.md
```diff
@@ -0,0 +1,457 @@
...
+For example, Toradex, a manufacturer of IoT boards, uses OSTree to deliver
+updates of their operating system TorizonCore to their devices. Toradex has also
+adopted the Uptane standard which implements TUF for automotive and other
+embedded usecases. In essence, Toradex's deployment implements the semantics
+described in this TAP--OSTree's content addressability is used for artifact
+integrity verification.
+
+# Specification
+
+The TUF specification uses file hashes in a number of contexts. All Targets
```
JustinCappos, line 97:
```suggestion
The TUF specification uses file hashes in a number of contexts. All targets
```

tap19.md
```diff
@@ -0,0 +1,457 @@
...
+the identifier is a hash calculated using an ecosystem-specific representation
+of the artifact. For examples, see the [motivating examples](#motivation).
+
+TUF can directly use these identifiers in its metadata instead of requiring
+users to calculate separate hash values. As TUF's metadata is agnostic to the
+hashing routine employed, this change does not require a change to the schema of
+how hashes are recorded. That said, TUF metadata will need to be updated to
+indicate the ecosystem in question.
+
+Presently, each entry in TUF's targets metadata has two key parts--the
```
JustinCappos, line 117:
```suggestion
Presently, each entry in TUF's Targets metadata has two key parts--the
```

tap19.md
```diff
@@ -0,0 +1,457 @@
...
+generalizable to all version control systems (VCSs). Consider Subversion (SVN),
+an alternative to Git. Like Git, SVN has a concept of recording changes which it
+calls _revisions_. However, SVN **does not** use a content addressed store for
+these revisions. Instead, each revision is identified by an auto-incrementing
+integer, one more than the previous revision. This identifier does not make any
+claims about the specific changes in the revision. Indeed, the identifier is
+entirely disconnected from the contents of the changes contained in the
+corresponding revision, and using it as a self certified value of a revision as
+prescribed in this document for Git is **dangerous**, entirely undermining the
+security properties offered by TUF.
```
JustinCappos, line 256: Isn't Git also dangerous since SHA1 is broken?

tap19.md
```diff
@@ -0,0 +1,457 @@
...
+
+Implementers must be therefore very careful with adopting this TAP for a new
+system. They must be familiar with the characteristic properties of content
+addressable systems. If they are implementing this TAP for an existing system
+they do not directly control, they must thoroughly and regularly
+[audit the system](#auditing-ecosystems-and-their-hash-computation-routines)
+mechanisms used by the system.
```
JustinCappos, line 263: It sounds here like you are saying this should not be a core part of TUF.  In other words, not all implementations should include content addressable support.  

We've been having some discussions about having optional parts of the TUF spec in TUF community meetings (raised by Jussi and @lukpueh if I recall correctly).  I think this case shows the rationale for this well.

POUFs/TAF-POUF/pouf2.md
```diff
@@ -0,0 +1,116 @@
...
+
+Firstly, while TUF identifies individual targets by their location relative to
+the mirror’s base URL, this POUF uses a URI to identify the specific Git
+namespace in a target repository. This URI is also used to locate the repository
+itself. The URI must use `git` as the scheme, clearly indicating that the entry
+pertains to a Git object.
+
+For metadata of some repository, a Targets entry is expected to map to a
+specific Git branch or tag. For each entry, whether a branch or a tag, a hash
+value must be recorded that clearly identifies the expected commit at the
```
JustinCappos, line 101: Is this SHA1?

JustinCappos (commented):
I have a few minor concerns with this as stated in my comments.  I am supportive in general, but would like to hear more from other community members (especially Lukas, Jussi, and Marina).  

If we had a more definitive policy about having "core TUF" and "TUF extensions" this would be an easy, immediate approve from me as a TUF extension.

tap19.md
adityasaky: Correct, I suspect most implementations may only need to support one ecosystem apart from standard files-on-disk (that may be optional if the TUF metadata itself is also on the content addressed store in question). Does the TAP need to clarify anything regarding core / non-core TUF here?

tap19.md
```diff
@@ -0,0 +1,457 @@
...
+algorithm should always generate the same hash value. These properties matter to
+hash algorithms selected by TUF implementations performing artifact integrity
+checks and so they must also exist in the content addressed ecosystem.
+
+## Unavailable Resources
+
+The availability of targets in a content addressed context is no different from
+that of regular files. For the metadata to be signed, the specific object must
+be available from the corresponding source.
+
```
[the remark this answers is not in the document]
adityasaky: Could you clarify what sort of complexity you have in mind?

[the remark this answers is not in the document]
JustinCappos: The code to figure out what a hash for something like this now becomes part of the trusted computing base on clients.  It's a lot harder to feel comfortable that you don't have a buffer overflow, API usage error, etc. if you are connecting to some IPFS library instead of just using SHA256 on a file.

tap19.md
adityasaky: Not to the same extent. In one case we have no integrity guarantees whatsoever. My thinking is that yes SHA-1 is broken but finding a valid collision in the git object format where the payload length also has to match makes it a lot more secure than applying this TAP to a system with no checks whatsoever.

JustinCappos: This argument makes me uncomfortable.  I think if you are going to allow this case, you at least need to be very explicit and clear about the pros and cons here.

adityasaky:
```
As noted before, content addressable systems typically use cryptographic hashes
over the contents of artifacts. A system that is a legitimate candidate for this
TAP must be thoroughly audited to validate its hash computation routines and
artifact integrity checks. If the system in question uses a weaker hash
algorithm, Git still uses SHA-1 for example, the accompanying POUF must justify
the choice.  Developers are also urged to monitor the development of the
ecosystem itself to ensure the assumptions of strong artifact integrity
validations continue to hold.
```

Is it fair to expect the case-by-case justification to live in the POUF? I'm not sure the TAP should argue for TAF/git alone.

znewman01:
CC @sudo-bmitch who mentioned that OCI is a little weird—they wrap the content-addressed blobs in some metadata, then use *that* as the hash

(I know OCI isn't a primary use case here, but it could be an interesting one.)

sudo-bmitch:
The TL;DR on OCI is you have the following:

- A tag that points to a manifest (effectively a mutable symbolic link to a hash)
- A manifest which is an OCI json structure containing either a list of manifests or blob hashes (but not both)
- Blobs are any data you want, stored by hash

If a blob isn't referenced by a manifest, a registry will usually garbage collect it after some time. So to structure things in OCI you'd want to identify what the individual blobs need to be, manifests to reference those blobs, and what tags to use to locate the manifests. The sha256 hash of the blobs will be the same across different CAS implementations, but the hash of an OCI manifest will probably only exist in the OCI implementation (but it may be similar in concept to the Git directory listing and hash).

tap19.md
```diff
@@ -0,0 +1,472 @@
...
+corresponding revision, and using it as a self certified value of a revision as
+prescribed in this document for Git is **dangerous**, entirely undermining the
+security properties offered by TUF.
+
+Implementers must be therefore very careful with adopting this TAP for a new
+system. They must be familiar with the characteristic properties of content
+addressable systems. If they are implementing this TAP for an existing system
+they do not directly control, they must thoroughly and regularly
+[audit the system](#auditing-ecosystems-and-their-hash-computation-routines)
+mechanisms used by the system.
```
mnm678, line 273:
```suggestion
mechanisms used.
```

mnm678 (approved):
One nit, but this looks ready to merge as a draft

JustinCappos (approved):
Given the recent changes, I've happy to approve merging this as a draft.

adityasaky:
@sudo-bmitch thanks for the OCI specific information! This TAP should cleanly apply to OCI by using the digest of the "root" manifest. I'd prefer to add this use case in a separate PR though rather than in this one, so that we can explore this structure some more.

PR #157, opened 2022-07-22
Add TAP enabling a verifiable chain of TUF metadata versions

adityasaky:
Overview: The idea with this TAP is to provide a way to verifiably track back from the latest metadata and examine all previous legitimate versions of Snapshot metadata. This in turn acts as an archive of all the legitimate Targets metadata issued in the repository.

Discussion: I'm looking for overall feedback on this approach and any thoughts folks may have. It's a rather small TAP but enables in-band recording of historical metadata that is important for some auditability specific use cases.

### DIFF
--- a/metadata-archival-tap.md
+++ b/metadata-archival-tap.md
@@ -156,6 +156,15 @@ hash. This continues all the way to the current Snapshot role, and therefore,
 when looking back, it is only possible to claim some Snapshot metadata file was
 legitimate when starting from the current legitimate role.
 
+Validating some Targets metadata file entails following the snapshot chain until
+one is found that records the Targets file in question. Note that there are two
+similar but different questions here. The first is if all the Targets files in a
+given set of metadata were ever valid. If every file in the set is found in a
+Snapshot file part of the chain, then the answer is yes. The second question is
+if a given set of Targets metadata were ever valid at the _same time_. In this
+case, all the Targets files must be found in the same Snapshot file part of the
+chain starting from the current Snapshot metadata file.
+
 # Security Analysis
 
 The inherent ability to track prior Snapshot roles does not weaken any of TUF's
--- a/metadata-archival-tap.md
+++ b/metadata-archival-tap.md
@@ -165,6 +165,32 @@ if a given set of Targets metadata were ever valid at the _same time_. In this
 case, all the Targets files must be found in the same Snapshot file part of the
 chain starting from the current Snapshot metadata file.
 
+## Supporting Verification of Prior Repository State
+
+While following the chain back to an older Snapshot version establishes its
+legitimacy, using the Targets metadata at that historical state requires them to
+be verifiable. Verification of the metadata requires a record of the keys
+trusted to sign that version of the metadata. For these scenarios, in addition
+to recording an entry of the previous Snapshot role, each Snapshot role must
+also record the current Root role.
+
+As before, actually finding the piece of metadata that corresponds to the entry
+is an implementation detail. Such metadata must not be implicitly trusted
+either. Instead, the standard TUF workflow must be followed to validate the
+specified version of the Root role. After that version is verified, TUF's
+verification workflow can be employed to validate some target. This means that
+including the Root role merely provides a hint about the version of the Root to
+be used for some Snapshot role and the set of Targets metadata it recorded. The
+specified Root role is not directly trusted just because it was recorded in the
+Snapshot role.
+
+It is possible to guess the version of the Root role to use for some historic
+state of the repository, perhaps by correlating the keys used to sign the
+Snapshot role with what is specified in the Root role. However, this is not
+guaranteed to work as Targets keys may have been rotated while Snapshot keys
+remained the same, leading to ambiguity in which version of the Root role should
+be used.
+
 # Security Analysis
 
 The inherent ability to track prior Snapshot roles does not weaken any of TUF's
--- a/metadata-archival-tap.md
+++ b/metadata-archival-tap.md
@@ -25,9 +25,14 @@ examination of every valid version of every Targets metadata file.
 
 While TUF's verification workflow only cares about the latest set of metadata
 for each role, a record of all prior metadata can be useful to trace back to
-the initial state of the repository. For use cases where examining old Targets
-metadata is necessary, we can only verify the veracity of any old file by
-checking if it was also part of the valid Snapshot role at the time. However,
+the initial state of the repository. One example of a system that needs to verify
+old metadata is [The Archival Framework (TAF)](https://github.com/openlawlibrary/taf),
+whose main goal is to protect digital law. After laws are published, they need
+to be authenticable for decades or even centuries.
+
+For use cases where examining old Targets metadata is necessary,
+we can only verify the veracity of any old file by checking
+if it was also part of the valid Snapshot role at the time. However,
 it may be impossible to validate an older Snapshot role as its keys may have
 been rotated or revoked since. Finally, a verifiable record of Snapshot and
 Targets metadata is also useful to identify if some previous metadata file is
--- a/metadata-archival-tap.md
+++ b/metadata-archival-tap.md
@@ -233,6 +233,14 @@ hashes, are looked up in the Snapshot role and validated. As such, an extra
 entry in the Snapshot role recording the characteristics of the previous
 version of the Snapshot metadata has no effect on backwards compatibility.
 
+If a TUF client that implements this TAP (i.e., it can follow a chain of
+Snapshot metadata back to validate some Targets metadata file) but encounters a
+TUF repository that does not implement this TAP, the verification of some
+retired Targets metadata file will fail. The verifiability provided by this TAP
+of prior Targets metadata requires buy-in from the repository. However, the TUF
+client in this case should be able to perform all the other standard TUF
+verification workflows.
+
 # Augmented Reference Implementation
 
 None at the moment.
--- a/metadata-archival-tap.md
+++ b/metadata-archival-tap.md
@@ -1,0 +1,230 @@
+* TAP:
+* Title: Using TUF Semantics for Verifiable Metadata Archival
+* Version: 1
+* Last-Modified:
+* Author: Aditya Sirish A Yelgundhalli
+* Status:
+* Content-Type: text/markdown
+* Created: 14-Jun-2022
+
+# Abstract
+
+Currently, TUF does not provide a mechanism to examine previous Snapshot
+metadata files in a verifiable manner. That is, while individual
+implementations may archive and store all metadata files, it is not possible to
+verify through TUF that older metadata files were not modified, replaced,
+reordered, or deleted. These old metadata files are useful because, together,
+they are an archive of all the Targets metadata generated for a repository. This
+TAP proposes that new Snapshot files must record the previous Snapshot file,
+creating a traversable chain of all Snapshot metadata files that were ever valid
+for the repository. This way, at any point in time, the "current" Snapshot role
+can be used to work back to the original Snapshot role, essentially enabling the
+examination of every valid version of every Targets metadata file.
+
+# Motivation
+
+While TUF's verification workflow only cares about the latest set of metadata
+for each role, a record of all prior metadata can be useful to trace back to
+the initial state of the repository. For use cases where examining old Targets
+metadata is necessary, we can only verify the veracity of any old file by
+checking if it was also part of the valid Snapshot role at the time. However,
+it may be impossible to validate an older Snapshot role as its keys may have
+been rotated or revoked since. Finally, a verifiable record of Snapshot and
+Targets metadata is also useful to identify if some previous metadata file is
+missing. This is important in instances where persisting of all metadata
+versions is a key goal.
+
+## Identifying if a Previous Version of the Metadata is Missing or Replaced
+
+Some deployments of TUF have an additional requirement that all versions of
+metadata issued by the repository must be stored for some period of time or
+forever for auditability. While the storage aspect is primarily an
+implementation detail, the semantics described in this TAP can be used to
+leverage TUF to identify if a previous version is missing. Further, the same
+semantics can also be used to verify that a claimant to a previous version of a
+metadata file is legitimate and has not been replaced.
+
+# Rationale
+
+Currently, the Snapshot role records some information about the top level and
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
I did it. I think we informally tawked about this idea with Mercury...

metadata-archival-tap.md
```diff
@@ -0,0 +1,220 @@
...
+creating a traversable chain of all Snapshot metadata files that were ever valid
+for the repository. This way, at any point in time, the "current" Snapshot role
+can be used to work back to the original Snapshot role, essentially enabling the
+examination of every valid version of every Targets metadata file.
+
+# Motivation
+
+While TUF's verification workflow only cares about the latest set of metadata
+for each role, a record of all prior metadata can be useful to trace back to
+the initial state of the repository. For use cases where examining old Targets
```
mnm678, line 28: Could you spell out 1-2 of these use cases?

metadata-archival-tap.md
```diff
@@ -0,0 +1,220 @@
...
+
+Implementers must be careful to only use the latest Snapshot role--their
+ability to track prior Snapshot roles must not lead to those older Snapshot
+metadata actually being used during TUF verification.
+
+This should not be a major concern for current implementations. They typically
+have a unique reference to the current Snapshot role, typically as the
+top-level `snapshot.EXT` file.
+
+# Backwards Compatibility
```
mnm678, line 202: You might want to discuss a client that attempts to use a verifiable chain for an existing repo that has not implemented this TAP.

metadata-archival-tap.md
```diff
@@ -0,0 +1,220 @@
+* TAP:
+* Title: Using TUF Semantics for Metadata Archival
+* Version: 1
+* Last-Modified:
+* Author:
```
mnm678, line 5: don't forget to give yourself credit :)

mnm678 (commented):
I like this idea, but am unclear about when it would be useful in practice.

trishankatdatadog:
> I like this idea, but am unclear about when it would be useful in practice.

It would help detect forking attacks, but only in conjunction with something like an append-only log (e.g., Rekor, Bitcoin blockchain).

metadata-archival-tap.md
```diff
@@ -0,0 +1,230 @@
...
+
+Currently, the Snapshot role records some information about the top level and
+delegated Targets metadata files. This is useful because it provides information
+about the latest versions of all the Targets metadata present in the repository.
+When a new version of a Targets file is available, the Snapshot role must also
+be updated to record it. As such, at any point in time, there is one "current"
+snapshot of the repository, and it records all the currently applicable Targets
+metadata.
+
+Therefore, in order to archive all Targets metadata, it is sufficient to record
```
renatav, line 57: Given a set of Targets metadata files, how do we know which Snapshot we're looking for? In TAF we bind metadata files together using commits - so Targets metadata at some revision are validated using a Snapshot file at that revision. Is it sufficient to find an archived Snapshot file according to which this given set of Targets is valid?

metadata-archival-tap.md
```diff
@@ -0,0 +1,230 @@
...
+      },
+      ...
+    },
+    "version": 4
+  }
+}
+```
+
+While each version of the Snapshot metadata can identify its predecessor, the
+realities of locating the file is left to each TUF implementation. In the
```
renatav, line 139: It would be very inconvenient to us to have to have all versions of the Snapshot metadata file on disk at the same time, seeing that we've already reached version 1000+.  So this flexibility is quite important to us

metadata-archival-tap.md
```diff
@@ -0,0 +1,230 @@
...
+ability to track prior Snapshot roles must not lead to those older Snapshot
+metadata actually being used during TUF verification.
+
+This should not be a major concern for current implementations. They typically
+have a unique reference to the current Snapshot role, typically as the
+top-level `snapshot.EXT` file.
+
+# Backwards Compatibility
+
+This TAP has no impact on backwards compatibility. This is because TUF's update
```
renatav, line 214: Will it be possible to add this to already existing repositories? Meaning that snapshot version, say, 500, starts listing older snapshot files? Or does version 2 have to already contain information about version 1?

renatav (commented):
Thank you for putting this together! We (the TAF team) have not put too much thought into what would happen if snapshot keys got revoked and it makes a lot of sense for a currently valid snapshot file to store hashes of older ones. Just shared some thoughts that crossed my mind while reading this, hope they make sense.

metadata-archival-tap.md
trishankatdatadog: Yes, if you know the git commit that binds together the Snapshot metadata with all of your Targets metadata, then you can use that git commit to find that Snapshot metadata, IIUC.

metadata-archival-tap.md
```diff
@@ -0,0 +1,230 @@
...
+be recorded, but instead, adds an extra detail in the Snapshot role, i.e., a
+record of the previous Snapshot role.
+
+That said, it is vital to consider some aspects of TUF's workings, and how they
+are impacted by the changes proposed in this TAP in greater detail.
+Specifically, since TUF metadata can expire or their keys may be revoked, it is
+important to consider what the meaning of validity of a since-replaced metadata
+file is when examined at some point in the future.
+
+## Archiving Revoked Snapshot Roles
```
trishankatdatadog, line 169: How do you know which Root metadata to use for old Snapshot metadata?

metadata-archival-tap.md
```diff
@@ -0,0 +1,230 @@
...
+a hash recorded in its successor.
+
+Implementers must be careful to start traversing the archival chain from the
+latest Snapshot role. For any pair of predecessor-successor metadata files, the
+predecessor's legitimacy in the future is based in the successor recording its
+hash. This continues all the way to the current Snapshot role, and therefore,
+when looking back, it is only possible to claim some Snapshot metadata file was
+legitimate when starting from the current legitimate role.
+
+# Security Analysis
```
trishankatdatadog, line 154: The TAP might also like to briefly discuss how [Transparent Logs](https://ssl.engineering.nyu.edu/blog/2020-02-03-transparent-logs) (e.g., gosumdb, Rekor) can be used to [detect](https://www.youtube.com/watch?v=PJ6b2eoq0NY&list=PLj6h78yzYM2NOCoaYcYbiAf4KPIF36T8t&index=12) forking attacks of the Snapshot metadata (i.e., that the repository "lied" about which Snapshot metadata was issued around that time).

metadata-archival-tap.md
renatav: That's how our system currently works. Should that be left to the implementers too?

metadata-archival-tap.md
adityasaky: Curious, do you still have all 1000+ snapshots backed up somewhere?

renatav: Just through git history

adityasaky: Ah right, of course.

metadata-archival-tap.md
renatav: Our team is working on a framework for securing digital law, relying on TUF, with one of it goals being archival authentication. Even 100 years in the future, we want people to be able to validate metadata files published today. This TAP points out the problem of validating old metadata following a keys rotation and sounds like something that we want to integrate into our system. Feel free to mention TAF here is that makes sense.

metadata-archival-tap.md
adityasaky: > Is it sufficient to find an archived Snapshot file according to which this given set of Targets is valid?

Yes, I think it is.

> Should that be left to the implementers too?

I'm wary of specifying too much here because it depends on how metadata is stored. In the general sense, following the chain back until a snapshot is found that records the wanted set of targets metadata is what this TAP is going for. I recognize that that's not optimal though. But implementation specific hints like the Git commit that binds them together should be fine. WDYT wrt specifying in spec?

metadata-archival-tap.md
adityasaky: @renatav could you add a description to that effect to the document?

metadata-archival-tap.md
adityasaky: Yes, this can start at any version. If you start it at version 500, you just can't trace back from the latest version, say 600, to the very first version. This impacts the claims you can make of those older metadata files.

metadata-archival-tap.md
adityasaky: Do you mean to identify which keys to use to verify a particular old Snapshot metadata?

metadata-archival-tap.md
renatav: I don't think it's necessary to provide any implementation specific hints. Stating that it is sufficient to find an archived Snapshot file according to which this given set of Targets is valid (by following the chain back to that Snapshot) would be a good addition imo.

metadata-archival-tap.md
renatav: @adityasaky Sure. Just a couple of sentence that sum up our problems and goals?

adityasaky: Yes, I think that'd do the trick. Thanks!

metadata-archival-tap.md
trishankatdatadog: Yes: how do you know which root metadata file is associated with which snapshot metadata files?

metadata-archival-tap.md
adityasaky: e98316f

Thanks @renatav!

metadata-archival-tap.md
trishankatdatadog: The chain back from the _current_ to whatever _previous_ snapsht metadata gives me some guarantees it's trustworthy, but not completely, because what if I want to go deeper and actually verify targets in there? How do I know which keys to use? That's why clients would need to know which root keys to use. Does this mean that snapshot should list the root again? @mnm678 @JustinCappos

metadata-archival-tap.md
adityasaky:
```
Validating some Targets metadata file entails following the snapshot chain until
one is found that records the Targets file in question. Note that there are two
similar but different questions here. The first is if all the Targets files in a
given set of metadata were ever valid. If every file in the set is found in a
Snapshot file part of the chain, then the answer is yes. The second question is
if a given set of Targets metadata were ever valid at the _same time_. In this
case, all the Targets files must be found in the same Snapshot file part of the
chain starting from the current Snapshot metadata file.
```

I added this to the end of the specification section. Does it help?

metadata-archival-tap.md
adityasaky: Done!

metadata-archival-tap.md
adityasaky: If there are usecases for this, recording the root should be fine. We should more strongly emphasize that the chain shouldn't include illegitimate snapshot metadata files signed after a key compromise in this case.

adityasaky: (Conversation with @mnm678) Perhaps the snapshot could hint at the root by storing only the version number rather than the hash of the root. The snapshot role shouldn't be trusted to unambiguously identify the root metadata.

trishankatdatadog: Yes, version number is fine. Now, another problem is: should clients trust the legitimacy of the snapshot chain (and hence the root metadata), or how can they verify that a snapshot is pointing to the right, unexpired root?

metadata-archival-tap.md
```diff
@@ -0,0 +1,263 @@
...
+ability to track prior Snapshot roles must not lead to those older Snapshot
+metadata actually being used during TUF verification.
+
+This should not be a major concern for current implementations. They typically
+have a unique reference to the current Snapshot role, typically as the
+top-level `snapshot.EXT` file.
+
+# Backwards Compatibility
+
+This TAP has no impact on backwards compatibility. This is because TUF's update
```
jku, line 239: The specification defines the snapshot "meta" contents as  _top-level targets metadata and delegated targets metadata [filenames]_. I wouldn't expect practical incompatibilities considering how it is used but strictly speaking this does not feel compatible.

metadata-archival-tap.md
```diff
@@ -0,0 +1,263 @@
...
+
+Second, by making prior states verifiable, we gain the ability to check if one
+or more Targets metadata files were ever valid, concurrently or otherwise.
+Essentially, this property establishes a verifiable archive of all Targets
+metadata issued by the TUF repository. One example of a system that needs to
+verify old metadata is
+[The Archival Framework (TAF)](https://github.com/openlawlibrary/taf), whose
+main goal is to protect digital law. After laws are published, they need to be
+authenticable for decades or even centuries. As such, the semantics described in
+this TAP enable the auditability of all TUF metadata issued by the repository.
```
jku, line 44: Couldn't this level of auditability could be reached by just storing all versions of metadata that are created? What is the advantage of this proposal in comparison?

jku (commented):
Overall this feels like adding complexity to the spec for fairly small benefit:
* Trishank already brought up transparency logs: if you _really_ need to know when things were made available, something like that sounds more useful
* for storing and auditing valid repository states, just archiving all metadata versions seems like a simpler solution (obviously you need a bit of software to validate a specific old repository version, but that should not be a massive job)

lukpueh (commented):
I don't quite understand the purpose of the root role in this proposal. If I ask snapshot to tell me what key it has been signed with, doesn't snapshot become its own (much weaker) root of trust? Even if I walk the chain of root metadata until the version snapshot hints at, that version of root could have been rendered meaningless by a subsequent version of root. I think the freshness of root metadata is indeed crucial.

As an example: snapshot v1 tells me that it has been signed with keys in root v1, but there already is a root v2, which tells me to not trust any keys from root v1 anymore. Now what's the value of verifying any signatures with keys from root v1?

As for chaining snapshots, can't we just use the already existent chaining properties of TUF metadata? That is: strictly incrementing timestamps - it doesn't even have to be every timestamp, only the ones that bump snapshot --, where each timestamp is the root of a consistent snapshot tree of metadata.

adityasaky:
Thanks for the reviews everyone! We've been discussing it and think that there are several valid points here re whether this should utilize mechanisms built into TUF or not. For now, I'm going to close this PR and withdraw this TAP. We'll focus on its bigger sibling #156 and re-think if we need the mechanisms described in this one.

PR #158, opened 2022-07-26
Tap14 updates

mnm678:
I updated TAP 14 based on some implementation questions from @abs007.

This replaces references to the now-rejected TAP 5 with TAP 13 and adds a new optional file to support a range of filesystem types.

### DIFF
--- a/tap14.md
+++ b/tap14.md
@@ -163,7 +163,7 @@ with a directory for each supported TUF specification version. These directories
 contain metadata that supports the given TUF specification version. Using these
 directories, a client is able to choose the most recent metadata they support.
 More details about this directory structure are contained in the [specification](#how-a-repository-updates).
-This TAP will also add a `supported-versions` field to root metadata so that the client
+This TAP will also add a `supported_versions` field to root metadata so that the client
 can discover the supported specification versions on the repository.
 
 On the client side, this TAP also requires maintenance of multiple versions that
@@ -250,7 +250,7 @@ has been updated to include the TUF Version field.
 
 Repositories will add two things. They MUST put metadata for new TUF
 specification versions in new
-directories. In addition, root metadata on a repository SHOULD add a `supported-versions` field
+directories. In addition, root metadata on a repository SHOULD add a `supported_versions` field
 to indicate which specification versions are supported.
 
 As described in the [Rationale](#rationale), repositories should support multiple
@@ -370,7 +370,7 @@ upgrades from version 1.x.y to version 2.0.0 may look like:
 Not all TUF repositories have a mechanism that is able to list all directories
 in a folder (the equivalent of the `ls` command). For these repositories (such
 as OCI registries or http servers), as well as to prevent specification version
-freeze attacks, root metadata SHOULD include a `supported-versions` field that lists
+freeze attacks, root metadata SHOULD include a `supported_versions` field that lists
 all versions supported by the repository to allow for client discovery. As it is
 included in root metadata, this field will be signed by a threshold of root keys,
 and will contain the following fields:
@@ -431,7 +431,7 @@ following procedure:
 
 * The client determines the latest version available on the repository by
 looking for the directory with the largest version number, or by parsing the
-`supported-versions` field in root metadata.
+`supported_versions` field in root metadata.
 * If the latest version on the repository is lower than the previous
 specification version the client used from this repository, the client
 should report an error and terminate the update.
@@ -565,7 +565,7 @@ A downgrade attack on the TUF specification version may be possible if an
 attacker is able to block access to a directory on the repository. This would
 mean that a client would use metadata from a previous specification version when
 performing an update.
-However, a client would be able to detect this attack using the `supported-versions`
+However, a client would be able to detect this attack using the `supported_versions`
 field listed in root metadata and signed with root keys.
 
 Clients that have previously used a repository will store the specification
--- a/tap14.md
+++ b/tap14.md
@@ -190,7 +190,7 @@ metadata files while allowing flexibility in how repositories and delegations
 are managed.
 
 For example, during the course of an update a client could use 4.0.0_parse_root,
-4.0.0_parse_snapshot, 4.0.0_parse_timestamp, 4.0.0_parse_targets,
+4.0.0_parse_timestamp, 4.0.0_parse_snapshot, 4.0.0_parse_targets,
 4.0.0_parse_delegation, and 3.0.0_parse_delegation.
 
 
--- a/tap14.md
+++ b/tap14.md
@@ -478,25 +478,25 @@ provided by each repository. Note that different TUF versions may use different
 hashing algorithms. If this is the case, both hashes should be verified
 independently.
 
-### TAP 5
+### TAP 13
 
-A TUF client that supports TAP 5 may download top-level metadata from multiple
-sources. However, breaking changes may require that multiple top-level metadata
-files use the same TUF specification version. To manage this, a root metadata
-file that includes a list of urls for metadata as described in TAP 5 should
-only include links to metadata files using the same TUF specification version as
-the root metadata file.
+A TUF client that supports TAP 13 may download targets metadata from multiple
+sources. However, breaking changes may require that all top-level metadata
+files use the same TUF specification version. To manage this, a client
+configuration file that lists a custom top-level targets metadata file
+as described in TAP 13 should only include targets metadata files using the
+same TUF specification version as the root metadata file.
 
-The owner of root metadata is responsible for ensuring that all metadata files
-linked in the root metadata use the same TUF specification version as the root
+The owner of the client configuration is responsible for ensuring that any
+custom targets metadata files use the same TUF specification version as the root
 metadata. If the repository that contains the root metadata supports multiple
-TUF specification versions, each root metadata file should list top-level
-metadata corresponding with its TUF specification version.
+TUF specification versions, the client configuration should list a targets metadata
+file that is available in all supported specification versions.
 
 Clients should verify that all top-level metadata use the same TUF specification
-version as the root metadata. If a metadata file linked in root metadata does
-not meet this criteria, it will be considered invalid and the client should
-use the next listed metadata file or terminate the update.
+version as the root metadata. If a metadata file provided by the TAP 13 configuration
+does not meet this criteria, it will be considered invalid and the client should
+terminate the update.
 
 ### Delegations
 
--- a/tap14.md
+++ b/tap14.md
@@ -1,7 +1,7 @@
 * TAP: 14
 * Title: Managing TUF Versions
 * Version: 1
-* Last-Modified: 07-August-2020
+* Last-Modified: 07-October-2022
 * Author: Marina Moore, Justin Cappos
 * Status: Draft
 * Content-Type: text/markdown
@@ -189,9 +189,9 @@ specification versions. This compromise allows TUF specification changes to affe
 metadata files while allowing flexibility in how repositories and delegations
 are managed.
 
-For example, during the course of an update a client could use 4.0.0_parse_root,
-4.0.0_parse_timestamp, 4.0.0_parse_snapshot, 4.0.0_parse_targets,
-4.0.0_parse_delegation, and 3.0.0_parse_delegation.
+For example, during the course of an update a client could use `4.0.0_parse_root`,
+`4.0.0_parse_snapshot`, `4.0.0_parse_timestamp`, `4.0.0_parse_targets`,
+`4.0.0_parse_delegation`, and `3.0.0_parse_delegation`.
 
 
 # Specification
@@ -250,7 +250,7 @@ has been updated to include the TUF Version field.
 
 Repositories will add two things. They MUST put metadata for new TUF
 specification versions in new
-directories. In addition, root metadata on a repository SHOULD add a `supported_versions` field
+directories. In addition, root metadata on a repository MUST add a `supported_versions` field
 to indicate which specification versions are supported.
 
 As described in the [Rationale](#rationale), repositories should support multiple
@@ -265,7 +265,7 @@ the directory, the repository creates and signs root, snapshot, timestamp, and
 top-level targets metadata using the new TUF specification version and places these
 metadata files in the directory. The root file should be signed by both the new
 root key and the current root key (the root key from the most recent metadata in
-the previous major specification version). Clients will now be able to use the new
+the previous major specification version). The new supported version number, and information about the new root metadata will then be added to `supported-versions` in all previously supported specification versions. Clients will now be able to use the new
 metadata files once their TUF specification versions are also updated. After an update to
 version 2.0.0, the repository structure may look like:
 
@@ -343,16 +343,6 @@ And the "signed" portion of root will include:
 
 where `BECOMES_OBSOLETE` is a timestamp.
 
-A repository should generate all TUF metadata, including root metadata, for all
-TUF versions that the repository supports. Any update should be reflected across
-all of these versions. For clarity, version numbers used for consistent
-snapshots should be consistent across all
-supported specification versions so that a client can find the current
-metadata files. This means that there may be a file at 1/3.root.json as well
-as 2/3.root.json. Root metadata files with the same consistent snapshot number must
-also use the same keys so that a client can find the next root metadata file in whichever
-specification version they support.
-
 For existing TUF clients to continue operation after this TAP is implemented,
 repositories may store metadata from before TUF 2.0.0 in the top-level
 repository (with no directory named 1). This allows existing clients to
@@ -367,19 +357,31 @@ upgrades from version 1.x.y to version 2.0.0 may look like:
   |- 2.x.y metadata files
 ```
 
-Not all TUF repositories have a mechanism that is able to list all directories
+In order to fascilitate the upgrade to a new specification version, root metadata for all supported spec versions MUST add a `supported_versions` field before upgrading to spec version 2.0.0.
+This field will serve two purpose.
+
+
+First, not all TUF repositories have a mechanism that is able to list all directories
 in a folder (the equivalent of the `ls` command). For these repositories (such
 as OCI registries or http servers), as well as to prevent specification version
-freeze attacks, root metadata SHOULD include a `supported_versions` field that lists
-all versions supported by the repository to allow for client discovery. As it is
-included in root metadata, this field will be signed by a threshold of root keys,
-and will contain the following fields:
+freeze attacks, the `supported_versions` field will lists
+all versions supported by the repository to allow for client discovery.
+
+Second, the root metadata format or versioning scheme may change between major specification versions.
+The `supported_versions` field will contain the filename and a secure hash of the first root metadata file that should be used in the new specification version.
+
+As it is
+included in root metadata, the contents of `supported_versions` will be signed by a threshold of root keys, securing against a specification version rollback, and signing the new trusted root metadata.
+This field will contain the following:
 
 ```
 { ...,
   "supported_versions" : [
     { "version": MAJOR_VERSION,
-      "path": FOLDER_NAME},
+      "path": FOLDER_NAME,
+      "root-filename": ROOT_FILENAME,
+      "root-digest": ROOT_DIGEST
+      },
     ...
   ]
 }
@@ -392,7 +394,11 @@ this supported major version (e.g. { "version": 2, "path": "2" }).
 `FOLDER_NAME` MUST NOT contain any subdirectories.
 In most cases, `MAJOR_VERSION` should match `FOLDER_NAME`.
 For backwards compatability, version 1 should be assumed to be in the top-level
-repository with no directory named 1.
+repository with no directory named 1. `ROOT_FILENAME` is the name of the root metadata file in the new specification version. `ROOT_DIGEST` is the digest of the new root metadata file.
+
+A repository should generate all TUF metadata, including root metadata, for all
+TUF versions that the repository supports. Any update to TUF targets or delegations should be reflected across
+all of these versions.
 
 ## Changes to TUF clients
 
@@ -422,7 +428,7 @@ remove the higher version and force the client to use a lower versioned,
 potentially less secure specification version.
 
 
-## Changes to the update process
+### Changes to the update process
 
 When a TUF client downloads metadata from a repository, the client must
 determine which specification version to use for the download. To do this,
@@ -430,26 +436,28 @@ the client looks for the highest supported version on the repository using the
 following procedure:
 
 * The client determines the latest version available on the repository by
-looking for the directory with the largest version number, or by parsing the
-`supported_versions` field in root metadata.
-* If the latest version on the repository is lower than the previous
+parsing the `supported_versions` field in the currently trusted root metadata.
+* The client determines the specification version to use using the following:
+	* If the latest version on the repository is lower than the previous
 specification version the client used from this repository, the client
 should report an error and terminate the update.
-* If the latest version on the repository is equal to that of the client, it
-will use this directory to download metadata.
-* If the latest version pre-dates the client specification version, it may call functions
+	* If the latest version on the repository is equal to that of the client, it
+will use this version to download metadata.
+	* If the latest version pre-dates the client specification version, it may call functions
 from a previous client version to download the metadata. The client may support
 as many or as few versions as desired for the application. If the previous
 version is not available, the client shall report that an update can not be
 performed due to an old specification version on the repository.
-* If the latest version on the repository is higher than the client spec
+	* If the latest version on the repository is higher than the client spec
 version, the client should report to the user that it is not using the most up
 to date version, and then perform the update with the directory that corresponds with the latest
 client specification version, if available. If no such directory exists, the client
 terminates the update.
+* If the client will be using a higher specification version than on their previous update, they will update their trusted root metadata by downloading the metadata available in the new specification version's directory at the root metadata location specified in the currently trusted root metadata. The client will verify that the hash of this file matches the digest listed in the current root metadata. If so, the client will set their trusted root metadata to the downloaded file.
 
-Once the supported directory is determined, the client shall attempt the update
-using the metadata in this directory.
+Once the supported directory is determined and root metadata is updated, the client shall attempt the update
+using the metadata in this directory. Every time the trusted root metadata
+file is updated, the client should perform this check to determine the specification version to use going forward.
 
 For example, if a client has a specification version of 3.5 and a repository has
 directories for 2, 3, and 4, the client will report that spec
@@ -461,27 +469,6 @@ Alternatively, if the same client downloads metadata from a repository with
 directories 1 and 2, the client could download metadata from 2 using
 a 2.x version of the client. If 2.x is not supported by the client, the client
 will report that it is unable to perform an update.
-
-Once the supported directory is determined, the client must validate root
-metadata from this directory. If the currently trusted root file saved on the
-client uses a specification version other than the supported version, the client will
-look for the next root file. The next root file may have been generated using
-the supported version, or it may have been generated using a previous TUF
-version. The client will look for the root file first in the supported version,
-then move to the previous versions until the next root file is found, or until
-the currently trusted root file's version is reached. All root files must be
-verified using the major version of the TUF client that corresponds with the
-major version of the root file.
-
-So, if the currently trusted root file is named 4.root.json and uses version
-1.0.0 and the highest supported version is 3.0.0, the client will look for
-5.root.json first in 3, then 2, then 1. If this file is found, the
-client will look for 6.root.json using the same process. To facilitate this, the
-client should maintain functions to parse root files from previous spec
-versions. Note that the library that supports a given spec version may also use
-a different naming convention than VERSION.root.json to find the root metadata.
-If the client does not support the specification version of a root file, the
-client shall terminate the update and report the specification version mismatch.
 
 When a client validates root and targets metadata, they may check for the
 `becomes_obsolete` field. If this field is present, the client should do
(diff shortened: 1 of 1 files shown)


tap14.md
```diff
@@ -478,25 +497,25 @@ provided by each repository. Note that different TUF versions may use different
...
+### TAP 13
 
-A TUF client that supports TAP 5 may download top-level metadata from multiple
-sources. However, breaking changes may require that multiple top-level metadata
-files use the same TUF specification version. To manage this, a root metadata
-file that includes a list of urls for metadata as described in TAP 5 should
-only include links to metadata files using the same TUF specification version as
-the root metadata file.
+A TUF client that supports TAP 13 may download targets metadata from multiple
+sources. However, breaking changes may require that all top-level metadata
```
lukpueh, line 503: How is TAP 13 related to downloading targets metadata from multiple sources? AFAIU, TAP 13 requires 3rd parties to upload the custom top-level targets metadata to the central repository.

lukpueh (changes requested):
LGTM modulo one type and a request for clarification (see inline)

lukpueh:
I have an non-PR-related question about [the following sentence in the TAP](https://github.com/theupdateframework/taps/blame/b7f539a9c8292ee51b88d46ddcf308aac435beb8/tap14.md#L304-L306):

> This allows them to differentiate between metadata that is expired due to deprecation and metadata that is expired due to an attack.

What does "expired due to an attack" mean?

tap14.md
```diff
@@ -163,6 +163,8 @@ with a directory for each supported TUF specification version. These directories
 contain metadata that supports the given TUF specification version. Using these
 directories, a client is able to choose the most recent metadata they support.
 More details about this directory structure are contained in the [specification](#how-a-repository-updates).
+The repository may also maintain a `supported-version` file so that the client
```
znewman01, line 166: `supported-versions`?

tap14.md
```diff
@@ -361,6 +365,20 @@ upgrades from version 1.0.0 to version 2.0.0 may look like:
   |- 2.0.0 metadata files
 ```
 
+Not all TUF repositories have file systems that are able to list all directories
+in a folder (the equivalent of the `ls` command). For these repositories (such
+as OCI registries or html servers), the repository SHOULD include a
```
znewman01, line 370: s/html/HTTP/?

tap14.md
```diff
@@ -361,6 +365,20 @@ upgrades from version 1.0.0 to version 2.0.0 may look like:
   |- 2.0.0 metadata files
 ```
 
+Not all TUF repositories have file systems that are able to list all directories
```
znewman01, line 368: nit: it's less about the FS, and more about the mechanism by which the repository is served

tap14.md
znewman01: also: move the example parenthetical to the previous sentence?

tap14.md
```diff
@@ -361,6 +365,20 @@ upgrades from version 1.0.0 to version 2.0.0 may look like:
...
+`supported-versions` file in the top of this directory structure that lists
+all versions supported by the repository to allow for client discovery. This file
+will have the following fields:
+
+```
+{ "supported_versions" : [VERSION, ...],
+
+```
+
+where `VERSION` is the name of a supported version in the format `version-number.0.0` (i.e. 2.0.0).
```
znewman01, line 380: While we're messing with this TAP...I kinda hate that we use "2.0.0" to mean "any 2.x.y version", I find it confusing...can we replace with just "2"?

(probably not in this PR)

znewman01 (approved):

tap14.md
lukpueh: Had the same thought while re-reading the TAP

tap14.md
mnm678: I want to avoid a double parenthetical in the previous sentence. I can re-work this bit if it's a big deal.

tap14.md
mnm678: Done. In my head the `2.0.0` is more clear, but it seems that this isn't true for others :)

lukpueh: The name could also by 2.x.y. :D

tap14.md
znewman01: I just ask because when I ready "not all TUF repositories have .... in a folder (the equivalent of ...)" my brain was really desperate for an example.

Maybe I'd advocate for dropping the `ls` comparison:

> Some TUF repositories are distributed over channels which do not support enumerating the subdirectories of the repository (for example, OCI registries or HTTP servers). To facilitate client discovery of available major versions, the repository SHOULD ...

Not a huge deal either way

tap14.md
```diff
@@ -178,7 +180,8 @@ version. A specification change may rely on more than one metadata file (for
 example a change in the signing process would affect all metadata types), so
 using the same specification version for top-level metadata allows for these
 large changes to the specification. For information about how this relates to
-url pinning and TAP 5, see [Special Cases](#tap-5). However, delegated targets metadata may not be
+targets metadata pinning and TAP 13, see [Special Cases](#tap-13). However,
+delegated targets metadata may not be
```
joshuagl, line 184: is this prohibited, or just a possibility? if the latter, we may want to say "might not".

tap14.md
```diff
@@ -245,13 +249,13 @@ has been updated to include the TUF Version field.
 ## How a repository updates
 
 Repositories will add metadata for new TUF specification versions in new
-directories.
+directories and may maintain a `supported-versions` file.
```
joshuagl, line 252:
```suggestion
directories and MAY maintain a `supported-versions` file.
```
or even SHOULD?

tap14.md
```diff
@@ -418,13 +437,13 @@ Once the supported directory is determined, the client shall attempt the update
 using the metadata in this directory.
 
 For example, if a client has a specification version of 3.5 and a repository has
-directories for 2.0.0, 3.0.0, and 4.0.0, the client will report that spec
-version 4.x is available, then download metadata from the 3.0.0 directory. This reporting is
+directories for 2, 3, and 4, the client will report that spec
+version 4.x is available, then download metadata from the 3 directory. This reporting is
 up to the discretion of an implementer, but it should be used to encourage
```
joshuagl, line 442: nit: we typically refer to implementation as libraries implementing the spec, and TUF deployments as applications or integrations

tap14.md
joshuagl: We really should remove the FS model from the spec, perhaps documenting the API the FS tree provides 🤔 
ref https://github.com/theupdateframework/specification/issues/103

tap14.md
mnm678: Does the update look right here?

tap14.md
```diff
@@ -342,24 +346,38 @@ TUF versions that the repository supports. Any update should be reflected across
...
+`supported-versions` file in the top of this directory structure that lists
+all versions supported by the repository to allow for client discovery. This file
+will have the following fields:
+
 ```
+{ "supported_versions" : [VERSION, ...],
+
+```
+
+where `VERSION` is the name of a supported major version (i.e. 2).
```
znewman01, line 380: What type is `VERSION`?

Should we indicate "major version 1, but no subfolder for it" separately?

tap14.md
```diff
@@ -245,13 +249,13 @@ has been updated to include the TUF Version field.
 ## How a repository updates
 
 Repositories will add metadata for new TUF specification versions in new
-directories.
+directories and SHOULD maintain a `supported-versions` file.
```
znewman01, line 252: Nit: this is a little out-of-nowhere, maybe move it into the next paragraph? If you use this directory structure, you SHOULD ..."

tap14.md
```diff
@@ -342,24 +346,40 @@ TUF versions that the repository supports. Any update should be reflected across
...
+`supported-versions` file in the top of this directory structure that lists
+all versions supported by the repository to allow for client discovery. This file
+will have the following fields:
+
 ```
+{ "supported_versions" : [VERSION, ...],
+
+```
+
+where `VERSION` is the integer name of a supported major version (i.e. 2).
```
znewman01, line 380: For forwards-compat, I'd like to consider making version be an object with a "number" key:

```
{ "supported_versions" : [
    { "number": 1 },
    { "number": 2},
    ...
  ]}
```

This will allow users to express more sophisticated things, like "folder X is compatible with both 3 and 4". I don't know if we'll need that, but there are situations where you might want multiple neighboring repos for non-major-version reasons (e.g. deprecating the

We could also skip it for now and replace the format of "supported-versions.json". Should we version supported-versions.json then?

tap14.md
```diff
@@ -342,24 +346,40 @@ TUF versions that the repository supports. Any update should be reflected across
...
+For backwards compatability, version 1 should be assumed to be in the top-level
```
znewman01, line 381: 👍 , this will make implementations much simpler

tap14.md
```diff
@@ -342,24 +346,40 @@ TUF versions that the repository supports. Any update should be reflected across
...
-  |- 2.0.0 metadata files
+- 1.x.x metadata files
+- 2
+  |- 2.x.x metadata files
+```
+
+Not all TUF repositories have a mechanism that is  able to list all directories
+in a folder (the equivalent of the `ls` command). For these repositories (such
+as OCI registries or http servers), the repository SHOULD include a
+`supported-versions` file in the top of this directory structure that lists
```
znewman01, line 371: `suppoerted-versions.json`?

tap14.md
```diff
@@ -529,7 +551,9 @@ attacks and changes to the security model that should be discussed.
 A downgrade attack on the TUF specification version may be possible if an
 attacker is able to block access to a directory on the repository. This would
```
znewman01, line 552: We talked briefly about having the root keys sign supported-versions.json. That feels reasonable to me, though maybe complicated.

tap14.md
```diff
@@ -342,24 +348,47 @@ TUF versions that the repository supports. Any update should be reflected across
...
+{ "supported_versions" : [
+    { MAJOR_VERSION: FOLDER_NAME},
+    ...
+  ]
+}
+
+```
+
+where `MAJOR_VERSION` is the integer representing a supported major version (i.e. 2)
+and `FOLDER_NAME` is the integer representing the folder containing metadata for
```
znewman01, line 388: Integer? I'd expect string.

tap14.md
znewman01: You change this here but it's inconsistent now 😛 , there's a bunch of places you use this file name.

tap14.md
mnm678: I guess a string of an integer makes the most sense

tap14.md
```diff
@@ -342,24 +348,47 @@ TUF versions that the repository supports. Any update should be reflected across
...
 ```
+{ "supported_versions" : [
+    { MAJOR_VERSION: FOLDER_NAME},
+    ...
+  ]
+}
+
+```
+
+where `MAJOR_VERSION` is the integer representing a supported major version (i.e. 2)
```
znewman01, line 387: nit: e.g., not i.e.

znewman01: but real comment: it might be nicer if you give a full example: `{ 2 : "2" }`

znewman01, line 387: Oh, JSON keys can only be strings. I think you want `{ "version": 2, "path": "2" }`

tap14.md
```diff
@@ -342,24 +348,47 @@ TUF versions that the repository supports. Any update should be reflected across
...
+and `FOLDER_NAME` is the string representing the folder containing metadata for
```
znewman01, line 388: Can folder name have more than one part?

znewman01, line 388: Is there any way to trick a client into going out-of-repo by passing `../../.ssh/id_rsa` or something?

znewman01 (commented):
Ahh, sorry to be annoying. I just had a bunch of worries come up

tap14.md
znewman01: This is addressed now.

znewman01: I *think* this is addressed? What if we say that clients MUST check that there's only `[a-z0-9-_]` characters in there (and same server side)

mnm678: If this file is signed by root, I don't think the clients need an additional check for the folder name, but they could add one for defense in depth.

tap14.md
```diff
@@ -163,6 +163,8 @@ with a directory for each supported TUF specification version. These directories
 contain metadata that supports the given TUF specification version. Using these
 directories, a client is able to choose the most recent metadata they support.
 More details about this directory structure are contained in the [specification](#how-a-repository-updates).
+The repository may also maintain a `supported-versions.json` file so that the client
```
trishankatdatadog, line 166: Located where exactly? Perhaps in [`.well-known`](https://en.wikipedia.org/wiki/Well-known_URI)?

tap14.md
```diff
@@ -244,14 +248,16 @@ has been updated to include the TUF Version field.
 
 ## How a repository updates
 
-Repositories will add metadata for new TUF specification versions in new
-directories.
+Repositories will add two things. They MUST put metadata for new TUF
+specification versions in new
+directories. In addition, repositories SHOULD maintain a `supported-versions.json` file
```
trishankatdatadog, line 253: Isn't this redundant with the note above?

tap14.md
```diff
@@ -342,24 +348,50 @@ TUF versions that the repository supports. Any update should be reflected across
...
 also use the same keys so that a client can find the next root metadata file in whichever
 specification version they support.
 
 For existing TUF clients to continue operation after this TAP is implemented,
 repositories may store metadata from before TUF 2.0.0 in the top-level
-repository (with no directory named 1.0.0). This allows existing clients to
+repository (with no directory named 1). This allows existing clients to
 continue downloading metadata from the repository. So a TUF repository that
-upgrades from version 1.0.0 to version 2.0.0 may look like:
+upgrades from version 1.x.x to version 2.0.0 may look like:
```
trishankatdatadog, line 360:
```suggestion
upgrades from version 1.x.y to version 2.0.0 may look like:
```

tap14.md
```diff
@@ -342,24 +348,50 @@ TUF versions that the repository supports. Any update should be reflected across
...
 
 
 ```
 - Targets files
-- 1.0.0 metadata files
-- 2.0.0
-  |- 2.0.0 metadata files
+- 1.x.x metadata files
```
trishankatdatadog, line 365:
```suggestion
- 1.x.y metadata files
```

tap14.md
```diff
@@ -342,24 +348,50 @@ TUF versions that the repository supports. Any update should be reflected across
...
+- 2
+  |- 2.x.x metadata files
```
trishankatdatadog, line 367:
```suggestion
  |- 2.x.y metadata files
```

tap14.md
```diff
@@ -342,24 +348,50 @@ TUF versions that the repository supports. Any update should be reflected across
...
+```
+
+Not all TUF repositories have a mechanism that is  able to list all directories
```
trishankatdatadog, line 370:
```suggestion
Not all TUF repositories have a mechanism that is able to list all directories
```

tap14.md
```diff
@@ -342,24 +348,50 @@ TUF versions that the repository supports. Any update should be reflected across
...
+in a folder (the equivalent of the `ls` command). For these repositories (such
+as OCI registries or http servers), the repository SHOULD include a
+`supported-versions.json` file in the top of this directory structure that lists
```
trishankatdatadog, line 373: If we are going to discuss this here, why not just link to this section from all the references to it above?

More importantly, I'm not sure about this idea. Why add even more complexity, especially requiring signatures? Are there simpler workarounds?

trishankatdatadog (commented):
Thanks, Marina! My biggest concern is on why we need a signed `supported-versions.json` somewhere for repositories that don't support listing of what appear to files in what appear to be folders. Could we please consider simpler mechanisms?

znewman01:
> My biggest concern is on why we need a signed supported-versions.json somewhere for repositories that don't support listing of what appear to files in what appear to be folders. Could we please consider simpler mechanisms?

@trishankatdatadog I think that's *most* repositories. For instance, go-tuf *only* has built-in support for HTTP remotes, and IIRC python-tuf is the same, but HTTP doesn't have a (standard/uniform) way to list the contents of a directory. We ran into this when attempting to build out TAP-14 support in python-tuf.

We're open to alternatives suggestions, of course. The first thought was to stick an `index.txt` file in each folder, but if there's a lot of root metadata files that gets annoying really quickly. So then we decided to filter it down to just a list of folders containing supported versions, but v1.x.y is a funky special-case (do we call it `1/` or `/`) so we figured being explicit was better.

Signing is something that feels appropriate to me as a "better safe than sorry" thing. I think we could get away with *not* requiring it, but it feels nicer to me to have the roots sign off on something as consequential as a major version transition.

tap14.md
mnm678: The signatures are to prevent a potential spec version freeze attack. This might not be the biggest concern as spec version upgrades should be rare, and once the client upgrades they can choose to not support the previous spec version, so happy to debate the signature part (especially as DoS attacks more generally are out of scope).

But users in some repos (like ones that use https) will need some mechanism to find the list of folders.

tap14.md
mnm678: I was thinking the metadata directory

tap14.md
mnm678: Yeah, I was trying to sign post what's in this section, but it does repeat the info from above. I can remove here if you think it isn't needed.

trishankatdatadog:
> @trishankatdatadog I think that's _most_ repositories. For instance, go-tuf _only_ has built-in support for HTTP remotes, and IIRC python-tuf is the same, but HTTP doesn't have a (standard/uniform) way to list the contents of a directory. We ran into this when attempting to build out TAP-14 support in python-tuf.

I see, I understand this better now, thanks.

Why Not Just ™️ use a convention to create and find the supported versions? For example, TAP 14 could mandate that metadata for each major spec version should go into its own directory (e.g., `1`, `2`, `3`...). Then, similar to the way we use [the N+1 rule](https://theupdateframework.github.io/specification/v1.0.31/index.html#update-root) to try to find the next root metadata, a client could also use the N+1 rule to try to find the latest supported major version. WDYT?

joshuagl:
> > @trishankatdatadog I think that's _most_ repositories. For instance, go-tuf _only_ has built-in support for HTTP remotes, and IIRC python-tuf is the same, but HTTP doesn't have a (standard/uniform) way to list the contents of a directory. We ran into this when attempting to build out TAP-14 support in python-tuf.
> 
> I see, I understand this better now, thanks.
> 
> Why Not Just ™️ use a convention to create and find the supported versions? For example, TAP 14 could mandate that metadata for each major spec version should go into its own directory (e.g., `1`, `2`, `3`...). Then, similar to the way we use [the N+1 rule](https://theupdateframework.github.io/specification/v1.0.31/index.html#update-root) to try to find the next root metadata, a client could also use the N+1 rule to try to find the latest supported major version. WDYT?

I'd guess because we want the spec to move away from assuming a filesystem to describe the API. That is, it should be possible to implement TUF without a filesystem even if today that requires interpreting rather than implementing certain portions of the spec.

trishankatdatadog:
> I'd guess because we want the spec to move away from assuming a filesystem to describe the API. That is, it should be possible to implement TUF without a filesystem even if today that requires interpreting rather than implementing certain portions of the spec.

I don't see why the try-and-see heuristic requires a filesystem tho?

trishankatdatadog:
> It's mostly because you can't actually fetch a directory in http, you have to try to fetch a file within the directory, so you have to know the name of the file. If a spec change changed the name of `root.json` or something, we couldn't rely on the existence of any particular file within the directory.

Could you please explain the use case for listing "files" in a "directory"? I'm still missing why this is needed.

znewman01:
TAP-14 puts different major-versioned metadata in directories like `2/` and `3/`. Clients are supposed to figure out what the latest supported major version is. In a filesystem, this is easy: `ls /repo/metadata/` and look for the directories.

Over HTTP, I can't `GET /repo/metadata/` (or rather I can, but what it gives me is totally server-dependent and may not include the list of contained files.

The next thing you might try is `GET /repo/metadata/2/root.json` etc.

There's three issues:

1. A major version change *may not* actually require a new `root.json`. (We could make this a requirement, but it's not in the TAP).
2. `root.json` in the next major version might be `root.yaml` or something. We want to be able to tell if there's a major version *beyond* the latest one supported in this client, so we can issue a warning (that's part of the existing TAP-14 spec).
3.  You may in fact have a repo that *skips* major versions: it might have been published under v1, and didn't get around to an upgrade until v3. So the N+1 trick doesn't work. (We could *require* in the TAP that you never skip major versions, but as-written you can't).

Beyond that, the implementation is pretty annoying. The supported-versions file is just a hint. You could make it a `.txt` file with 1 directory per line, but the fact that v1 is in the root complicates that a little bit, so we decided to be more specific and clear in the data format.

trishankatdatadog:
> Beyond that, the implementation is pretty annoying. The supported-versions file is just a hint. You could make it a `.txt` file with 1 directory per line, but the fact that v1 is in the root complicates that a little bit, so we decided to be more specific and clear in the data format.

I see, that's very helpful, thanks! I'm just slightly worried about the relative complexity this adds while providing relatively little benefit. I'm not convinced that there is a use case where clients will want to choose which major version they want to support. It would certainly be a very complicated client, and I can't imagine a high demand for such a feature. My guess is that people will simply use different clients for different major metadata versions of the same repo.

So, sorry, I'm not trying to be difficult here, but I think we can punt this feature to the future, if it's ever needed. WDYT?

znewman01:
>  I'm just slightly worried about the relative complexity this adds while providing relatively little benefit. I'm not convinced that there is a use case where clients will want to choose which major version they want to support. It would certainly be a very complicated client, and I can't imagine a high demand for such a feature. My guess is that people will simply use different clients for different major metadata versions of the same repo.
>
> So, sorry, I'm not trying to be difficult here, but I think we can punt this feature to the future, if it's ever needed. WDYT?

This seems like an argument against TAP 14 more generally, rather than this change in particular, right?

If anything, this change makes the client-side implementation substantially easier. It shouldn't be particularly hard on the repo side either.

mnm678:
> I see, that's very helpful, thanks! I'm just slightly worried about the relative complexity this adds while providing relatively little benefit. I'm not convinced that there is a use case where clients will want to choose which major version they want to support. It would certainly be a very complicated client, and I can't imagine a high demand for such a feature. My guess is that people will simply use different clients for different major metadata versions of the same repo.
> 
> So, sorry, I'm not trying to be difficult here, but I think we can punt this feature to the future, if it's ever needed. WDYT?

The proposal here is very similar to using a different client for different major versions, but the TAP 14 checks provide a way for clients to make the decision about which of these versions to use when potentially interacting with multiple repositories. The TUF spec has to account for all types of TUF implementations, including repos with multiple clients, clients with multiple repos, and combinations thereof, which is why this TAP was designed to cover all of these cases.

Implementations could choose to only list a single supported version at a time as either a client or repo, and that will limit their implementation to that particular version.

The reason we need this now-ish is to support future compatibility with breaking changes (like DSSE). The sooner we solve breaking changes the sooner we can move forward with a bunch of other TAPs that require some amount of non-backwards compatibility, without breaking things for our many exisiting users.

trishankatdatadog:
I see, thanks for the explanation, Zach and Marina! Ok, now I'm convinced that we do need the contents of `supported-versions.json` somewhere. However, I'm not convinced that we need it as a separate file, which raises tricky questions such as who signs it, does it expire, etc. If it's expected to be signed by the root role, why not just stick it in there as an optional field in every root metadata for a major spec version? A little redundant, but simplifies life overall in my opinion. WDYT?

trishankatdatadog:
> I don't think it's expected to be signed by the root role. Don't think of it as a metadata file, think of it as a hint.

SGTM. Now the Q is: do we sign or not sign this hint?

JustinCappos:
FYI all, @Marina Moore ***@***.***> and I had a nice chat about this
issue and laid out the pros and cons.

She's going to capture this in writing here so we have it for posterity.
Let's wait for her analysis and then discuss further.

mnm678:
Can we get consensus on this approach from TAP maintainers? @JustinCappos and @trishankatdatadog have already looked. @lukpueh and @joshuagl wdyt?

tap14.md
```diff
@@ -369,14 +369,15 @@ upgrades from version 1.x.y to version 2.0.0 may look like:
 
 Not all TUF repositories have a mechanism that is able to list all directories
 in a folder (the equivalent of the `ls` command). For these repositories (such
-as OCI registries or http servers), the repository SHOULD include a
-`supported-versions.json` file in the top of this directory structure that lists
-all versions supported by the repository to allow for client discovery. This file
-will be signed by a threshold of root keys, and the `signed` portion will have
-the following fields:
+as OCI registries or http servers), as well as to prevent specification version
+freeze attacks, root metadata SHOULD include a `supported-versions` field that lists
```
joshuagl, line 373: Observation, this feels like it will be a good place to provide additional guidance for repository operators on managing "supported" versions.

joshuagl (dismissed):
Adding a field to root metadata feels like a more appropriate path forwards. 👍

lukpueh:
Unrelated observation [about this paragraph](https://github.com/theupdateframework/taps/blame/0b27e8849c1c4bb5095ba16d68e8b1db5ed2c7c5/tap14.md#L192-L194):
> For example, during the course of an update a client could use 4.0.0_parse_root, 4.0.0_parse_snapshot, 4.0.0_parse_timestamp, 4.0.0_parse_targets, 4.0.0_parse_delegation, and 3.0.0_parse_delegation.

If the order of function calls should represent the course of an update, `*_parse_timestamp` should be before `*_parse_snapshot`.

lukpueh:
Unrelated observation [about this paragraph](https://github.com/theupdateframework/taps/blame/0b27e8849c1c4bb5095ba16d68e8b1db5ed2c7c5/tap14.md#L227-L228):

> If the change adds a new feature that is backwards compatible, for example in TAP 4 and TAP 10 [...]

TAP 10 [claims](https://github.com/theupdateframework/taps/blob/master/tap10.md#backwards-compatibility) that it is backwards incompatible.

lukpueh:
Ping https://github.com/theupdateframework/taps/pull/158#issuecomment-1203886869, cc @mnm678

tap14.md
```diff
@@ -441,7 +475,7 @@ major version of the root file.
 
 So, if the currently trusted root file is named 4.root.json and uses version
 1.0.0 and the highest supported version is 3.0.0, the client will look for
-5.root.json first in 3.0.0, then 2.0.0, then 1.0.0. If this file is found, the
+5.root.json first in 3, then 2, then 1. If this file is found, the
```
lukpueh, line 478: This process requires a `VERSION.root.json` filename convention, but in discussion [above](https://github.com/theupdateframework/taps/pull/158#issuecomment-1261277925) and in the [google doc](https://docs.google.com/document/d/1QcjEKzL0dHMVDmpkeIZuNIkN6VzduKBHa-4Mw-XSSx4/edit?usp=sharing) suggest that root may change its filename anytime.

tap14.md
```diff
@@ -397,7 +430,8 @@ the client looks for the highest supported version on the repository using the
 following procedure:
 
 * The client determines the latest version available on the repository by
-looking for the directory with the largest version number.
+looking for the directory with the largest version number, or by parsing the
+`supported-versions` field in root metadata.
```
lukpueh, line 434: I'm trying to wrap my head around the upgrade+update process. Would someone be willing to walk me through an extremely simple example for the following scenario:
- repo has 1.root.json with supported version 2
- client has same 1.root.json as trusted root, and client library supports versions 2 and 3
- repo upgrades to spec version 3 and updates root to version 2
- client upgrades+updates root using the `supported_versions` field

Questions:
- Which files need to be created in the repo during the upgrade/update and which keys are used to sign them?
- Which files are downloaded by the client and in what order?

lukpueh (dismissed):
Following a chat with @mnm678 and @JustinCappos, I just pushed a couple of minimal fixes for the concerns I raised. @joshuagl, can you green-light again?

tap14.md
JustinCappos: Glad to have the chat.  Let's make sure this is clarified in writing for posterity too...

tap14.md
```diff
@@ -397,7 +430,8 @@ the client looks for the highest supported version on the repository using the
 following procedure:
 
 * The client determines the latest version available on the repository by
-looking for the directory with the largest version number.
+looking for the directory with the largest version number, or by parsing the
```
trishankatdatadog, line 433: WDYM "looking for the..."? Do you mean guess-and-try?

tap14.md
```diff
@@ -441,10 +475,12 @@ major version of the root file.
 
 So, if the currently trusted root file is named 4.root.json and uses version
 1.0.0 and the highest supported version is 3.0.0, the client will look for
```
abs007, line 477: Should it be explicitly mentioned in the example that the min supported major version is 1 because it might be confusing to interpret that. In this case it makes sense since 1 is the first version by default but someone might not be able to figure that out in their first read. What do you think?

JustinCappos: I'd like to second this.  We should be explicit that TUF 1.0 is /1/, and that the major version number of the specification will be the directory number (2.0 -> /2/, 3.0 -> /3/, etc.) moving forward.

tap14.md
joshuagl: Rather than have a client always try to do the "look for version=N+1", wouldn't it make more sense for the TAP to suggest _only_ using the `supported_versions` field from the root metadata?

mnm678:
Based on discussion with @JustinCappos and @lukpueh I simplified the root metadata update mechanism in this TAP. The new mechanism allows the root version numbers to diverge between specification versions without compromising the security of an upgrade. I will update the PoC shortly to reflect this change.

JustinCappos (approved):
Thanks for the hash / supported_versions change to the text.  Looks ready to me.

trishankatdatadog (approved):
LGTM to merge as draft for now, please do address my comments as we discussed earlier:
* The unclear benefits of `becomes_obsolete` at the expense of added complexity (e.g., delegatees could accidentally create conflicts)
* If we keep `becomes_obsolete`, clarify whether it uses the same datetime format as `expires`
* Clarify that root rotation is done as before this TAP for the current known MAJOR spec version before looking at (the latest) `supported_versions`
* Clarify whether a root metadata file refer to MAJOR versions lower than or equal to the current `spec_version` in that file (and metadata version directory)
* There are only 3 cases for checking version numbers: <, =, or >. The current text is ambiguous about the < case. Please clarify.

PR #162, opened 2022-10-04
Move Tap14 to deferred with some additional clarification of repository operations

mnm678:
Requires #158

Please ignore all but the last 3 commits, this is already rebased onto #158 to prevent conflicts. I'll clean up the history here once that is merged.

This pr adds some requested clarification of repository operations, and proposes moving this TAP to accepted based on the PoC in python-tuf.

### DIFF
--- a/tap14.md
+++ b/tap14.md
@@ -266,11 +266,13 @@ parent directory to save space. In this case the metadata files (in directories 
 version) can point to target files relative to the parent directory. After creating
 the directory, the repository creates and signs root, snapshot, timestamp, and
 top-level targets metadata using the new TUF specification version and places these
-metadata files in the directory. The root file should be signed by both the new
-root key and the current root key (the root key from the most recent metadata in
-the previous major specification version). The new supported version number, and information about the new root metadata will then be added to `supported_versions` in all previously supported specification versions. Clients will now be able to use the new
-metadata files once their TUF specification versions are also updated. After an update to
-version 2.0.0, the repository structure may look like:
+metadata files in the directory. The root file should be signed by the new root keys
+listed in the file. A digest of the signed root metadata, along with the new
+supported version numver  will then be added to the `supported_versions` field in
+a new root metadata file for all previously supported specififcation versions.
+Clients will now be able to use the new metadata files once their TUF specification
+versions are also updated. After an update to version 2.0.0, the repository
+structure may look like:
 
 
 ```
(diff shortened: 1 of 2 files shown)


JustinCappos (changes requested):
FYI: the main index of TAPs also needs to be updated to change the status to accepted...

JustinCappos (dismissed):

lukpueh (changes requested):
Some comments, which I think should be addressed before accepting the TAP

- This PR has conflicts.

- The TAP does not point to a prototype implementation, and the existing prototype implementation (https://github.com/theupdateframework/python-tuf/pull/2114) does not represent the current state of the TAP yet.

- The recently merged PR [#158](https://github.com/theupdateframework/taps/pull/158) had a few spelling mistakes and typos in the last commit (e.g. `supported-versions` with dash instead of underscore)

- The `supported_versions.path` field seems to be redundant with the convention that spec version `X.Y.Z`  metadata lives in an `X` directory.

- How does `root-filename` evolve over root versions? E.g. `2/2.root.json` and `2/20.root.json` can't both say `{"version": 3, "root-filename": "1.root.json", ...}`, because `3/1.root.json` needs to be signed by *"the root key from the most recent metadata [at the time of creating the new root (??)] in the previous major specification version"* and keys may have changed between `2/2.root.json` and `2/20.root.json`. 

  Does this mean that root updates need to be synchronized across multiple versions and `root-filename` needs to be bumped on each root update? IMO that seems necessary for clients, in order to update spec version coming from different root versions. But it also makes the different spec versions less autonomous. Either way I think the TAP should clarify this in the "repository updates" section.

  Besides, how does `root-filename` evolve for the own and previous versions? E.g. if a repo supports spec versions 2, 3 and 4, what would the `supported_versions` entries for 2 and 3 likely look like in `3/8.root.json`?

- I'd generally like the client update process to be more algorithmic also with regards to the existing client workflow. E.g. @trishankatdatadog [suggests](https://github.com/theupdateframework/taps/pull/158#pullrequestreview-1138209759) that the client should rotate root before looking at `supported_versions`. Is that the case? The TAP does not make that clear.

mnm678:
@lukpueh I addressed your text concerns, and will be updating the prototype implementation soon. Hopefully the implementation will further clarify the root update process.

mnm678:
The implementation now reflects the changes in this TAP. @lukpueh @JustinCappos this is ready for review

tap14.md
```diff
@@ -357,48 +413,24 @@ upgrades from version 1.x.y to version 2.0.0 may look like:
...
-this supported major version (e.g. { "version": 2, "path": "2" }).
-`FOLDER_NAME` MUST NOT contain any subdirectories.
-In most cases, `MAJOR_VERSION` should match `FOLDER_NAME`.
-For backwards compatability, version 1 should be assumed to be in the top-level
-repository with no directory named 1. `ROOT_FILENAME` is the name of the root metadata file in the new specification version. `ROOT_DIGEST` is the digest of the new root metadata file.
-
-A repository should generate all TUF metadata, including root metadata, for all
-TUF versions that the repository supports. Any update to TUF targets or delegations should be reflected across
-all of these versions.
+For additional optimizations, the repository could any duplicate metadata files
```
JustinCappos, line 430: grammar error here in reference to 'any'

JustinCappos (changes requested):
I looked and the recent changes are fine with me.  (Modulo a minor edit below.)

JustinCappos (dismissed):
I approve

jkjell (approved):

JustinCappos (approved):
Moving to deferred.  Lukas and others will add other changes.

lukpueh (approved):

PR #165, opened 2023-01-19
[TAP 8] Interoperability with Fulcio TAP

mnm678:
Adds a section to clarify the interaction between TAP 8 and the Fulcio TAP.

### DIFF
--- a/tap8.md
+++ b/tap8.md
@@ -290,6 +290,8 @@ If this happens, the attacker could use TAP 8 to create large-scale rotations
 to null to perform a denial of service. However, this attack can be safely
 recovered from using the delegator to replace all effected keys. This is
 especially effective if the top-level targets is controlled by offline keys.
+For this reason, we recommend that both root and top-level targets are
+controlled by an offline, non-Fulcio key.
 Further, an attacker with access to a large number of identities through
 a compromised Fulcio server or OIDC issuer would be able to perform a
 similar attack without TAP 8 by signing invalid or attacker-controlled
--- a/tap8.md
+++ b/tap8.md
@@ -277,9 +277,9 @@ directly trust this team), the foo.rotate files can be safely removed.
 
 ## Interoperability with Fulcio TAP
 
-TAP 8 can be used with the Fulcio identities proposed in the Fulcio TAP.
+TAP 8 can be used with the Fulcio identities proposed in [TAP 18](https://github.com/theupdateframework/taps/blob/master/tap18.md).
 To do so, the `sigstore-oidc` keytype may be used in the rotate file
-as a KEY. Using this extension, TAP 8 can be used to rotate Fulcio identities
+as a KEY and this keytype may also be used to sign the rotate file. Using this extension, TAP 8 can be used to rotate Fulcio identities
 as well as other key types.
 
 The interaction between the Fulcio TAP and this TAP may impact the security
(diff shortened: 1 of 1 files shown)


tap8.md
```diff
@@ -275,6 +275,13 @@ directly done to Alice, Bob, Charlie and Dan with a threshold of 2
 (e.g. the delegation expired, and the person in charge decided to
 directly trust this team), the foo.rotate files can be safely removed.
 
+## Interoperability with Fulcio TAP
+
+TAP 8 can be used with the Fulcio identities proposed in the Fulcio TAP.
+To do so, the `sigstore-oidc` keytype may be used in the rotate file
+as a KEY. Using this extension, TAP 8 can be used to rotate Fulcio identities
+as well as other key types.
```
JustinCappos, line 283: Is it really this simple? How do you understand the ordering of multiple rotate files in this case?  What are the trust assumptions here?  Is it okay if a newer signed rotate file in Rekor is trusted over an older one because of the order of upload?

mnm678: I added the below paragraph about security considerations, and added a mention of signing here

tap8.md
```diff
@@ -275,6 +275,26 @@ directly done to Alice, Bob, Charlie and Dan with a threshold of 2
 (e.g. the delegation expired, and the person in charge decided to
 directly trust this team), the foo.rotate files can be safely removed.
 
+## Interoperability with Fulcio TAP
+
+TAP 8 can be used with the Fulcio identities proposed in [TAP 18](https://github.com/theupdateframework/taps/blob/master/tap18.md).
```
joshuagl, line 280: Nit: probably just use a relative link so that we can follow it to the sibling TAP, regardless of where/how we are browsing the repo?

```suggestion
TAP 8 can be used with the Fulcio identities proposed in [TAP 18](tap18.md).
```

tap8.md
```diff
@@ -275,6 +275,26 @@ directly done to Alice, Bob, Charlie and Dan with a threshold of 2
...
+To do so, the `sigstore-oidc` keytype may be used in the rotate file
+as a KEY and this keytype may also be used to sign the rotate file. Using this extension, TAP 8 can be used to rotate Fulcio identities
+as well as other key types.
```
joshuagl, line 283: nit: wrapping
```suggestion
as a KEY and this keytype may also be used to sign the rotate file. Using this
extension, TAP 8 can be used to rotate Fulcio identities as well as other key
types.
```

tap8.md
```diff
@@ -275,6 +275,26 @@ directly done to Alice, Bob, Charlie and Dan with a threshold of 2
...
+
+The interaction between the Fulcio TAP and this TAP may impact the security
+considerations in this TAP. If the Fulcio server or OIDC issuer is compromised,
+an attacker could gain control over all TUF roles controlled by that entity.
+If this happens, the attacker could use TAP 8 to create large-scale rotations
+to null to perform a denial of service. However, this attack can be safely
+recovered from using the delegator to replace all effected keys. This is
+especially effective if the top-level targets is controlled by offline keys.
```
joshuagl, line 291: Perhaps we should explicitly recommend that top-level targets is an offline, non Fulcio/TAP 18, key?

joshuagl (approved):

JustinCappos (approved):

Issue #37, opened 2017-05-31
Acceptance of TAP 8: Key rotation and explicit self-revocation


vladimir-v-diaz:
TAP 8 proposal is available at https://github.com/theupdateframework/taps/blob/master/tap8.md

Pull requests related to TAP 8:
https://github.com/theupdateframework/taps/pull/20

Discussion of TAP 8 on the TUF mailing list:
https://groups.google.com/d/msg/theupdateframework/Yj1z3PkMRZc/n8wXM4VZBgAJ

vladimir-v-diaz:
Hi, @hannesm.

I came across an old comment of yours on the issue [discussion for TAP 3 (multi-role delegations)](https://github.com/theupdateframework/taps/issues/32#issuecomment-318095539).
In the comment you say that a naming mechanism is still needed for TAP 8 (key rotation).  Have you found a solution yet?  Is there anything else blocking TAP 8?

(I'm just checking in on the status of TAP 8)

hannesm:
It looks like GitHub ate my email; yes, I'm working on a revision of TAP8 - unfortunately not yet done, but hopefully by the end of this week.  I'll open a PR then and let's discuss in that how to move forward.  Sorry for the delay.

vladimir-v-diaz:
No problem!

Thanks, and we're looking forward to TAP 8 revision.

hannesm:
@vladimir-v-diaz due to other projects, my implementation of tuf + tap8 is only progressing slowly (only doing this in my spare time, working on other projects for funding).  so far I have basic tuf nearly finished (in OCaml), and will work over the summer on finalising tuf+tap8.  I'll get back to tap8 once implemented, sorry for the delay.

awwad:
I think this is a good and wise feature add. I was reviewing @mnm678's [TUF PR draft implementing TAP 8](https://github.com/theupdateframework/tuf/pull/783) and combed through TAP 8 and chatted with Marina a bit. I think that there are some conceptual issues and a few clarifications that should be made. I'm going to stick to only the two bigger conceptual points for this post:

### Root Key Rotation Efficiency Improvement
In the course of the client's traversing the path from the currently trusted root file to the most recent root file, I'm not yet convinced that it's okay to only fetch root files that include key/threshold changes for the root keys:
[...]

PR #167, opened 2023-01-27
[TAP 8] Simplify rotate file names

mnm678:
Simplify rotate files per the discussion in the [related issue](https://github.com/theupdateframework/taps/issues/37#issuecomment-1402443495).

### DIFF
--- a/tap8.md
+++ b/tap8.md
@@ -148,7 +148,7 @@ signatures wrapper as in tuf spec, not shown here):
 ```python
 {
     "_type" : "rotate" ,
-    "previous" : PREV_FILENAME
+    "version" : VERSION
     "role" : ROLE,
     "keys" : {
         KEYID : KEY
@@ -159,9 +159,10 @@ signatures wrapper as in tuf spec, not shown here):
 
 Where ROLE, KEYID, KEY, and THRESHOLD are as defined in the original
 tuf spec.  The value of ROLE has to be the same as the role for the
-delegation.  The value of THRESHOLD is its new value.  PREV_FILENAME is
-the name of the previous rotate file in the chain, or the empty string if this is
-the first rotate file for this role.  The keys specify the new valid keys
+delegation.  The value of THRESHOLD is its new value.  VERSION is
+the integer version number of rotate files for this role. Version
+numbers MUST increase by exactly 1 from the previous rotate file for
+this role. The keys specify the new valid keys
 and associated key ids (which may be a subset or superset of
 the old ones).  A rotate file does _not_ contain an expiration date,
 it is meant to be signed once and never modified.  The rotate
@@ -176,20 +177,13 @@ Let's consider a motivating example, project foo is delegated to Alice.
 Alice's computer with the key material got stolen, but the disk was
 encrypted.  To be safe, she decides to get her key from her backup and
 roll over her key to a fresh one.  To do that, she creates a file
-`foo.rotate.ID.PREV`.  This file contains her new key, a threshold of 1, and
+`foo.rotate.VERSION`.  This file contains her new key, a threshold of 1, and
 is signed with her old key.  She signs her targets file with the new key,
 and uploads both the rotate file and the freshly signed targets file to
 the repository.
 
-The first filename suffix, referred as `ID` above and below, is the hex
-representation of the SHA256 hash of the concatenation (using "." as
-separator) of the KEYIDs (in ascending lexical order), and the old
-threshold value, encoded decimal as ASCII (0x31 for 1, 0x32 for 2,
-0x31 0x30 for 10).
-
-The second suffix, referred to as 'PREV' is a SHA256 hash of the previous
-rotate file, or the empty string if there is no previous rotate file.
-This prevents rotation cycles.
+The first filename suffix, referred as `VERSION` above and below, is the
+same as the VERSION listed in the metadata.
 
 The existing delegation to Alice's old key is still valid. The client will
 start with this delegation and look for rotation files to determine the current
@@ -198,16 +192,15 @@ set of trusted keys.
 If the delegation is changed to include her new key, it will also be valid. Any
 old rotate files for this role should be deleted and removed from snapshot on
 the next snapshot key rotation. The client will determine the correct rotate file
-to begin the chain using the set of keys in the delegation for the role as the `ID`.
+by starting from VERSION 0.
 
 ## Client workflow
 
 A client who wants to install foo now fetches Alice's targets file, and
-during verification looks for a file named `foo.rotate.ID.PREV` in the
-rotate folder, ID and PREV are explained above using Alice's old keyid.
-The client sees the file, fetches
+during verification looks for a file named `foo.rotate.0` in the
+rotate folder. The client sees the file, fetches
 it and verifies this rotate file using the public key from the delegation.
-The client then looks for a rotate file with the new keyid, repeating until
+The client then looks for a rotate file with version `1`, repeating until
 there is no matching rotate file to ensure up to date key information. This
 establishes trust in Alice's new key, and the client can now verify the
 signature of Alice's targets file using the new key.  If key data is missing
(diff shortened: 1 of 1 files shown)


tap8.md
```diff
@@ -159,37 +159,37 @@ signatures wrapper as in tuf spec, not shown here):
 
 Where ROLE, KEYID, KEY, and THRESHOLD are as defined in the original
 tuf spec.  The value of ROLE has to be the same as the role for the
-delegation.  The value of THRESHOLD is its new value.  PREV_FILENAME is
-the name of the previous rotate file in the chain, or the empty string if this is
-the first rotate file for this role.  The keys specify the new valid keys
+delegation.  The value of THRESHOLD is its new value.  VERSION is
+the integer version number of rotate files for this role. Version
+numbers MUST increase by exactly 1 from the previous rotate file for
```
JustinCappos, line 164: How does this interact with multiple repositories?  If Alice rotates a key, she must effectively rotate it everywhere, since anyone can download her rotate file from elsewhere and then add it somewhere else.

tap8.md
```diff
@@ -233,25 +232,20 @@ Multi-role delegations are handled using the same methodology.
 Let's consider the project baz, initially delegated to a multi-role
 threshold (of 2) to roles foo, and bar, each of which have a threshold of 2
 keys.  When they want to add a keyid from Dan to the foo role, the current foo
-keyholders create a foo.rotate.ID.PREV file, where ID is as described above (the
-SHA256 of the concatenated key ids of foo, and the character 0x32). This contains
-all previous foo keys, as well as Dan's key and a new threshold.  PREV is the
-SHA256 of "" as this is the first rotation for foo.  The file
-foo.rotate.ID.PREV is signed by at least 2 current keyids of foo.
+keyholders create a foo.rotate.0 file. This contains
```
JustinCappos, line 235: Why is this version 0 instead of 1?

JustinCappos (changes requested):
Needs some help / clarification in a few places...

tap8.md
mnm678: This is the first rotate file, so it starts at 0

tap8.md
mnm678: I added a section about the interaction with TAP 4.

tap8.md
jku: all other role versions start from 1

tap8.md
```diff
@@ -198,16 +198,15 @@ set of trusted keys.
...
-rotate folder, ID and PREV are explained above using Alice's old keyid.
-The client sees the file, fetches
+during verification looks for a file named `foo.rotate.0` in the
+rotate folder. The client sees the file, fetches
 it and verifies this rotate file using the public key from the delegation.
-The client then looks for a rotate file with the new keyid, repeating until
+The client then looks for a rotate file with version `1`, repeating until
 there is no matching rotate file to ensure up to date key information. This
 establishes trust in Alice's new key, and the client can now verify the
 signature of Alice's targets file using the new key.  If key data is missing
```
jku, line 212: Earlier it is said 

> Clients MUST use snapshot metadata to ensure that they recieve all rotate files in the chain.

but then client workflow here only talks about (my interpretation) downloading each rotate version from 0 and continuing until 404. 

1. I don't understand how snapshot data is used.
2. The workflow for existing (cached) rotate files must be defined much better: I can sort of imagine how it's supposed to work but the model is completely new to TUF client implementations so really needs to be clear (as clients might have foo.rotate.0 cached from an earlier delegation: it used to be a valid rotate file but is not valid anymore -- now the same file may contain completely different content that may now be valid. This pretty much never happens to any role in a current spec repository)

hannesm:
Thanks for this proposal. This indeed simplifies the proposal a lot.

Now, thinking about "why did we use a hash initially", as far as I remember:
- consider a setup where something is delegated to a team (i.e. a quorum of 2 from alice, bob, carla)
- how can this team modify its validity? (i.e. "we want a quorum of 3 from alice, bob, carla, doreen")

Would this still be possible with this simplification? From my memory, the reason to use  a hash of the validity expression came from the observation that there's no distinct file for a delegation.

But I've neither followed up closely with TUF development, nor am I certain that the scenario described above is worth considering in your use cases. For me, there is the question "who is part of a team?" and "where are signatures put?" -- and I want to minimize the amount of files that have the requirement to have multiple signatures (since that means the file has to be passed to multiple entities before being put (and being valid) into the repository).

tap8.md
mnm678: I updated the TAP to start at version 1 for consistency

tap8.md
mnm678: I updated this workflow to include snapshot and caching.

mnm678:
> Would this still be possible with this simplification? From my memory, the reason to use a hash of the validity expression came from the observation that there's no distinct file for a delegation.

Yes, the team could create a rotate file for the role with the next version number (so 1 to start). The goal here is to replicate the file name uniqueness from the hash with a version number. The rotate files are still signed with the previously trusted set of keys, and so only the existing team can create a valid rotate file.

tap8.md
```diff
@@ -198,60 +198,56 @@ set of trusted keys.
...
-there is no matching rotate file to ensure up to date key information. This
+A client who wants to install foo now fetches Alice's targets file and will determine the set of trusted keys for this role as follows. The client will start with the keys listed in the delegation as trusted keys for this role.
+The client will then look for all files that begin with `rotate/foo.rotate` in
+the snapshot metadata. The client will process these in version order (ie starting with `rotate/foo.rotate.1`, then `rotate/foo.rotate.2` by first checking for this version file in a local cache. The client will then fetch
+the rotate file from remote. If the remote file is a rotation to null, and is signed with the currently trusted keys, the client will halt the verification of this metadata and act as if it is unverified when continuing the update process. Otherwise, the client should ensure that the cached file is identical to the remote version. The client will then verify the rotate file using the currently trusted public key. If a rotate file is successfully verified, the client will update the set of trusted keys for this role to be the set listed in the rotate files. If key data is missing
```
JustinCappos, line 208: What does unverified mean here?  How do you handle cases like a threshold where some of the individual keys have been rotated?

tap8.md
```diff
@@ -198,60 +198,56 @@ set of trusted keys.
...
 repository.
 
 The root role should ensure that all previous rotate files are removed when
 it delegates to a new chain of trust. This saves space and simlifies the client
 search for rotate files.
 
+## Interoperability with TAP 4 (Multiple repository consensus)
+
+If multiple repositories use the same role definition, rotate files may need to be coordinated to ensure consistency. There are two cases when roles are used across multiple repositories, the first is a mirror of a repository, and the second is two repositories that have different contents, but share a particular targets role. In the case of a mirror, all rotate files should be copied, and the repository can perform as usual, including any rotations. In the second case, the repository manager must ensure that they have the same set of trusted keys for a role after all rotations. This can be achieved by copying all rotate files for a role or by creating a delegation to the final set of trusted keys indicated by rotate files on another repository. If the latter method is used, note that future rotate files may not by copied as the version numbers will not start from one.
```
JustinCappos, line 228: This is too up in the air now.  I think we need to figure out more of TAP 8 / rotation to definitively say this should be done.

JustinCappos (changes requested):

tap8.md
```diff
@@ -159,99 +159,117 @@ signatures wrapper as in tuf spec, not shown here):
...
-
-Timestamp and snapshot keys can rotate as well, leading to ROLE.rotate.ID.PREV
-files, where ID and PREV are as described above. Each file is signed by a quorum
-of old keys, and contains the new keys. If the timestamp key is renewed by
-the root, all timestamp.rotate files can be safely removed from the
-repository.
+A client who wants to install foo now fetches Alice's targets file and will
+determine the set of trusted keys for this role as follows. The client will start
+with the keys listed in the delegation as trusted keys for this role.
+The client will then look for all files that begin with `rotate/foo.rotate` in
```
JustinCappos, line 209: At some point we need to consider what to do with respect to naming more generally.  However, I think this is fine to merge.

JustinCappos (approved):

JustinCappos (approved):

joshuagl (approved):
I really like the simplification here, nice. I made one minor suggestion to clarify rotate file names, otherwise this looks great.

joshuagl (approved):

JustinCappos (approved):
This set of changes LGTM.  Other PRs need to merge before this TAP can move forward.

PR #141, opened 2021-07-27
TAP for TUF developer key management

mnm678:
This TAP proposes the use of Fulcio to improve developer key management in TUF.

PR #168, opened 2023-02-07
Fulcio draft

mnm678:
Requires #141 

Add a TAP number and make follow-up changes from #141

### DIFF
--- a/tap18.md
+++ b/tap18.md
@@ -1,7 +1,7 @@
-* TAP:
+* TAP: 18
 * Title: Ephemeral identity verification using sigstore's Fulcio for TUF developer key management
 * Version: 0
-* Last-Modified: 27/07/2021
+* Last-Modified: 07/02/2023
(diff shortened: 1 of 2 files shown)


tap18.md
```diff
@@ -52,7 +52,7 @@ A signature using a Fulcio key MUST include the Fulcio certificate for use in ve
 ```
 {
    "keyid" : KEYID,
-   "bundle": BUNDLE
+   "sig": BUNDLE
```
jku, line 55: This might still be problematic -- the spec does define sig as "hex encoded signature" so that would have to be changed to "key specific content".

That said, I think it might work in practice (as an example current python-tuf likely would work just accept this metadata) but the other option would  be to leave "sig" as is and just add a new (sigstore-signature specific) field for the bundle.

JustinCappos (approved):
I approve.  I sent minor editing comments to Marina over slack, but there is nothing factually wrong with the TAP proposal and I approve it being merged as Draft.

trishankatdatadog (approved):

mnm678:
Thanks! I will merge this to get the TAP number assigned, and follow up with a pr for the minor edits

Issue #33, opened 2017-05-31
Discussion of TAP 4: The map file


vladimir-v-diaz:
TAP 4 proposal is available at https://github.com/theupdateframework/taps/blob/master/tap4.md

Pull requests related to TAP 4:
https://github.com/theupdateframework/taps/pull/1
https://github.com/theupdateframework/taps/pull/2
https://github.com/theupdateframework/taps/pull/4
https://github.com/theupdateframework/taps/pull/5
https://github.com/theupdateframework/taps/pull/14
https://github.com/theupdateframework/taps/pull/15
https://github.com/theupdateframework/taps/pull/16

Discussion of TAP 4 on the TUF mailing list:
https://groups.google.com/d/msg/theupdateframework/gHOgkjxMpEU/mYw2CdxiBwAJ

vladimir-v-diaz:
We decided to make a major change to TAP 4 where the map file is no longer required.  Instead, the TAP will now advise adopters on how they can support multiple repository consensus on entrusted targets.
For example, it shows how they can deal with prioritized repositories, terminating repositories, etc. and uses the map file as an example of how the reference implementation supported multiple repository consensus.

TAP 4's revision can be followed here:
https://github.com/theupdateframework/taps/pull/64

vladimir-v-diaz:
@endophage

You requested that we make changes to TAP 4, namely to remove the map.json requirement.  Please let us know what you think of the [revised proposal](https://github.com/theupdateframework/taps/blob/master/tap4.md).

vladimir-v-diaz:
I'd like to document here some of your requested changes...

> The only true security feature I see in TAP4 is the matching against multiple repos. Is there a way to greatly simplify the tap to focus on that specific feature? That focus would make it much clearer what it's actually trying to achieve.
> 
> If the TAP only contained the client side aliasing and not the multi-repo consensus, it would have no security benefits and be purely client side UX sugar. Would that get accepted as a TAP? IMO that falls squarely in the space of letting clients decide how best to interact with users and doesn't belong in the spec.
> 
[...]

PR #174, opened 2023-03-22
Update the list of referenced implementations for TAP 4

rdimitrov:
The following PR updates the list of referenced implementations for TAP 4

Details:

* deletes the old entry for python-tuf as this is no longer applicable
* adds rdimitrov/go-tuf-metadata

### DIFF
--- a/tap4.md
+++ b/tap4.md
@@ -290,8 +290,12 @@ A TUF repository does not need to change in any way to support this TAP.
 
 # Augmented Reference Implementation
 
-Pull Request [#504](https://github.com/theupdateframework/python-tuf/pull/504)
-implements multiple repository consensus on entrusted targets via a map file.
+There's an implementation of TAP 4 under the `multirepo` package in 
+[rdimitrov/go-tuf-metadata](https://github.com/rdimitrov/go-tuf-metadata).
+It implements the multiple repository consensus on entrusted targets via a
+map.json file. The package provides a multi-repository client API for creating
+a client, searching for a given target and downloading it all by implementing
+the search algorithm described in TAP 4.
 
 # Copyright
 
--- a/tap4.md
+++ b/tap4.md
@@ -294,7 +294,7 @@ There's an implementation of TAP 4 under the `multirepo` package in
 [rdimitrov/go-tuf-metadata](https://github.com/rdimitrov/go-tuf-metadata).
 It implements the multiple repository consensus on entrusted targets via a
 map.json file. The package provides a multi-repository client API for creating
-a client, searching for a given target and downloading it all by implementing
+a client, searching for a given target and downloading it all by following
 the search algorithm described in TAP 4.
 
 # Copyright


mnm678 (approved):

trishankatdatadog (approved):

tap4.md
```diff
@@ -290,8 +290,12 @@ A TUF repository does not need to change in any way to support this TAP.
...
 # Augmented Reference Implementation
 
-Pull Request [#504](https://github.com/theupdateframework/python-tuf/pull/504)
-implements multiple repository consensus on entrusted targets via a map file.
+There's an implementation of TAP 4 under the `multirepo` package in 
+[rdimitrov/go-tuf-metadata](https://github.com/rdimitrov/go-tuf-metadata).
+It implements the multiple repository consensus on entrusted targets via a
+map.json file. The package provides a multi-repository client API for creating
+a client, searching for a given target and downloading it all by following
+the search algorithm described in TAP 4.
```
trishankatdatadog, line 298: I think the search algorithm described in the TAP is still a bit ambiguous TBH (e.g., how exactly do you met the threshold requirement --- do you go through an n-choose-k number of repos until you find an agreement?), but this is a good start 🙂 Curious to know whether you ran into these issues, too, and how you solved them.

rdimitrov: That's a very valid point, @trishankatdatadog! Thank you for bringing that up as I had planned to raise it on yesterday's TUF meeting but totally forgot! 

I've implemented it to the point that it looks for enough target infos that match the necessary threshold and I assume that the responsibility to better handle the mappings is up to the maintainer of the map.json file. 
What I mean is there's the case that if you have listed 5 repos for a single target and somehow you ended having 2/5 match one info and the other 3/5 match another AND your threshold is 2, there's no way to tell which group should be the trusted one. It's not perfect, but the maintainer of the map.json file should be more diligent around how these mappings can be exploited so they limit the possibility of such use cases.

As for how go-tuf-metadata will handle that, I find it the safest to error out in case such conflict happens where there is more than 1 group that matches the configured threshold as otherwise it may lead to an unknown behaviour. In any case, I'm open to discuss and find a better solution for that since it's indeed a potential point of error.

joshuagl: Should we open an issue to discuss, and resolve, the ambiguity of the algorithm?

tap4.md
```diff
@@ -290,8 +290,12 @@ A TUF repository does not need to change in any way to support this TAP.
 
 # Augmented Reference Implementation
 
-Pull Request [#504](https://github.com/theupdateframework/python-tuf/pull/504)
```
joshuagl, line 293: This PR was merged and shipped in legacy python-tuf, I wonder if it's worth still mentioning the [`MultiRepoUpdater`](https://github.com/theupdateframework/python-tuf/blob/febe6c3dc30195078a5487d68d7db708e7de8479/tuf/client/updater.py#L161) implementation in the TAP?

joshuagl (approved):

tap4.md
rdimitrov: Let's do that 👍 I'll go ahead and file one

tap4.md
rdimitrov: Yeah, I don't think it looks good to list something as a reference implementation while at the same time, it has been deprecated for a few years now

tap4.md
rdimitrov: It turns out this was already raised in https://github.com/theupdateframework/taps/issues/33 and should be discussed in https://github.com/theupdateframework/specification/issues/88. I guess there is no need to create another one, right?

tap4.md
joshuagl: Makes sense

PR #175, opened 2023-04-06
[TAP 18] add separate field to signature for the bundle

joshuagl:
Per previous discussion and the securesystemslib implementation this change updates TAP 18 to include an additional field in the signature format for explicit inclusion of the bundle.


lukpueh (approved):
Thanks!

PR #183, opened 2024-02-07
[TAP 8] Describe DoS attack and prevention

mnm678:


kipz:
@mnm678 it looks like you're thinking this is something that implementors can choose mitigate, but couldn't this be a formal part of the spec; that is, put the limit in the metadata and make it so the client MUST fail if the number of revocation files is greater than the limit?

JustinCappos:
I would tend to agree that at a minimum, the existence of a limit and discussion around this should be in the document.  

Note that if there is a limit, what if some clients have a different limit than others?  This will cause a divergence in views of the repo.  So this almost certainly needs to be a per-repository setting based upon information in the root metadata.

kipz:
Makes sense.

I think this differs from the similar attack mentioned in #37 whereby any key holder can upload several different versions of the same metadata file. A revocation limit is something that could be defined ahead of time, whereas the number of metadata file versions really is unknown.

mnm678:
I added some text about this. I think it can be addressed at the repository side by limiting uploads of rotate files after a certain point. This allows the repository to set the limit based on their available storage/hosting resources, and to share this limit with all clients.

tap8.md
```diff
@@ -208,6 +208,8 @@ old rotate files for this role should be deleted and removed from snapshot on
 the next snapshot key rotation. The client will determine the correct rotate file
 to begin the chain using the set of keys in the delegation for the role as the `ID`.
 
+The repository SHOULD set a limit to the number of rotate files per role. This limit should be clear to all key holders (for example, it could be in repository documentation or added to root metadata). Once this number of rotate files is reached, the repository will reject rotations for this role until a new delegation is made to it, clearing all previous rotations.
```
JustinCappos, line 211: So how would this work in practice?  Bob creates a new targets file signed with a new key, but with the same role name?  What if Alice and Charlie both delegate to Bob?  Do they need to coordinate their update to the delegation so both happen at the same time?

mnm678: This got me thinking about multiple-delegation more generally. To simplify this case, I updated the [other TAP 8 pr](https://github.com/theupdateframework/taps/pull/167) to discourage re-delegation to the same role name. With this change, after the rotate file limit is reached, new delegations should be to a new role name. I'll update this pr to align with those changes.

joshuagl (approved):

lukpueh (approved):

PR #185, opened 2024-03-21
Assign self-revocation tap a number: 20

joshuagl:
Assign the self-revocation tap a number and list it in the TAP index.


mnm678 (approved):

JustinCappos (approved):
LGTM

PR #186, opened 2024-03-26
Tap 20 Acceptance

mnm678:
Depends on #187 (accept TAP 8 first)

### DIFF
--- a/tap20.md
+++ b/tap20.md
@@ -131,7 +131,7 @@ where NULL is the null key.
 
 Rotation files are immutable unless replaced with a revocation.
 This is the only case in which they can be replaced or
-modified.  If a client wants to rotate to a different
+modified.  If a key holder wants to rotate to a different
 key, without having access to their currently delegated private key,
 this requires a key revocation by the delegating metadata.
 
--- a/tap20.md
+++ b/tap20.md
@@ -77,7 +77,7 @@ from that proposal to support key revocation.
 ## Rotate file (from TAP 8)
 
 The signed portion of a `rotate` file is as follows (there's also a
-signatures wrapper as in tuf spec, not shown here):
+signatures wrapper as in the TUF specification, not shown here):
 
 ```python
 {
@@ -92,7 +92,7 @@ signatures wrapper as in tuf spec, not shown here):
 ```
 
 Where ROLE, KEYID, KEY, and THRESHOLD are as defined in the original
-tuf spec.  The value of ROLE has to be the same as the role for the
+TUF specification.  The value of ROLE has to be the same as the role for the
 delegation.  The value of THRESHOLD is its new value. VERSION is the index of
 this rotate file, incrementing by 1 on each rotation. The keys specify the new
 valid keys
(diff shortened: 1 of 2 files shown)


tap20.md
```diff
@@ -125,23 +125,23 @@ where NULL is the null key.
...
 that many security-sensitive revocations that require key use by
 multiple parties, will now be much easier to do.
 
-Clients need to take care to check for rotation to a null key (rotate
+Clients need to take care to check for revocation (rotate
 files that contain a null key).  This shall be handled in the
 same manner as an invalid metadata signature on by the role possessing
-the key. The role will be invalid until it is re-delegated to with a new key.
+the key. The role will be invalid until it is re-delegated to with a new
+rolename and key.
```
jkjell, line 144: Sorry if this was covered during draft but, why can't the rolename be reused with a new key?

jkjell (commented):
Looks good to me! Just had one question.

tap20.md
mnm678: This is because of the discovery mechanism for rotate/revocation files. They are named `ROLENAME.rotate.VERSION`, so if the role is revoked then replaced the old rotate files might still be around (as they can only be deleted when snapshot is rotated to prevent a rollback), leading to confusion. Forcing a new rolename might make the names less human-readable, but otherwise should have minimal impact.

jkjell (approved):

lukpueh (commented):
This TAP doesn't really mention how it changes the client workflow, apart from "Clients need to check for rotations to a null key,". Is this vagueness intended? Or am I missing something?

Also, I think it uses the term "client" ambiguously.

mnm678:
> This TAP doesn't really mention how it changes the client workflow, apart from "Clients need to check for rotations to a null key,". Is this vagueness intended? Or am I missing something?
> 
> Also, I think it uses the term "client" ambiguously.

I pushed some updates that describe the client workflow. This mostly relies on TAP 8 for download of rotate files, and just adds a step of checking for a rotation to null.

"client" is meant to refer to a TUF client. Is there a way I can make this more clear in the text?

lukpueh:
> I pushed some updates that describe the client workflow. This mostly relies on TAP 8 for download of rotate files, and just adds a step of checking for a rotation to null.

In that case, we should accept TAP 8 first before we accept TAP 20. Seems a bit fishy otherwise. 

> "client" is meant to refer to a TUF client. Is there a way I can make this more clear in the text?

I think in the sentence "If a client wants to rotate to a different key, without having access to their currently delegated private key" client means something else.

mnm678:
> > I pushed some updates that describe the client workflow. This mostly relies on TAP 8 for download of rotate files, and just adds a step of checking for a rotation to null.
> 
> In that case, we should accept TAP 8 first before we accept TAP 20. Seems a bit fishy otherwise.

Fair enough, I'll update the description so that this pr relies on #187 

> 
> > "client" is meant to refer to a TUF client. Is there a way I can make this more clear in the text?
> 
> I think in the sentence "If a client wants to rotate to a different key, without having access to their currently delegated private key" client means something else.

got it, fixed that one

jkjell (approved):

PR #188, opened 2024-04-02
Accept Tap 16

mnm678:
This pr adds the rust implementation, makes a couple of simplifications to the specification section to ensure the efficient Merkle BPT implementation is supported, and moves the TAP to accepted.


trishankatdatadog (dismissed):

kipz (approved):
LGTM

trishankatdatadog (approved):

jkjell (approved):

JustinCappos:
Let's make sure the feature add into python-tuf can land first.  I do think
we'll do something transparency log related / adjacent as a TAP before we
merge this.  Then this can use that as a building block.

PR #189, opened 2024-11-25
DRAFT: TAP-21 - Scale-out Architecture for High-volume Repositories

ergonlogic:
Coming from theupdateframework/specification/issues/309, and following discussion at the TUF [community meeting on 2024-11-01](https://hackmd.io/jdAk9rmPSpOYUdstbIvbjw#November-1-2024-Meeting), we've drafted "TAP-21 - Scale-out Architecture for High-volume Repositories".

This TAP is still very much a "draft". Several sections are currently marked "TBD". Our motivation for a PR at this stage is to validate the motivation and rationale, as well as the calculations we undertook to better understand the issues at hand.

### DIFF
--- a/tap1.md
+++ b/tap1.md
@@ -1,14 +1,12 @@
----
-TAP: 1
-Title: TAP Purpose and Guidelines
-Version: 2
-Last-Modified: 30-Nov-2020
-Author: Trishank Karthik Kuppusamy, Lois Anne DeLong, Justin Cappos, Joshua Lock
-Status: Active
-Content-Type: text/markdown
-Created: 07-Sep-2016
-Post-History: 08-Sep-2016
----
+* TAP: 1
+* Title: TAP Purpose and Guidelines
+* Version: 2
+* Last-Modified: 30-Nov-2020
+* Author: Trishank Karthik Kuppusamy, Lois Anne DeLong, Justin Cappos, Joshua Lock
+* Status: Active
+* Content-Type: text/markdown
+* Created: 07-Sep-2016
+* Post-History: 08-Sep-2016
 
 # What is a TAP?
 
--- a/tap1.md
+++ b/tap1.md
@@ -1,12 +1,14 @@
-* TAP: 1
-* Title: TAP Purpose and Guidelines
-* Version: 2
-* Last-Modified: 30-Nov-2020
-* Author: Trishank Karthik Kuppusamy, Lois Anne DeLong, Justin Cappos, Joshua Lock
-* Status: Active
-* Content-Type: text/markdown
-* Created: 07-Sep-2016
-* Post-History: 08-Sep-2016
+---
+TAP: 1
+Title: TAP Purpose and Guidelines
+Version: 2
+Last-Modified: 30-Nov-2020
+Author: Trishank Karthik Kuppusamy, Lois Anne DeLong, Justin Cappos, Joshua Lock
+Status: Active
+Content-Type: text/markdown
+Created: 07-Sep-2016
+Post-History: 08-Sep-2016
+---
 
 # What is a TAP?
 
--- a/tap21.md
+++ b/tap21.md
@@ -1,0 +1,191 @@
+* TAP: 21
+* Title: Scale-out Architecture for High-volume Repositories
+* Version: 1
+* Last-Modified: 25-Nov-2024
+* Author: Christopher Gervais christopher@consensus.enterprises, Derek Laventure derek@consensus.enterprises
+* Type: Standardization
+* Status: Draft
+* Content-Type: text/markdown
+* Created: 25-Nov-2024
+* TUF-Version: 1
+* Post-History: TBD
+
+# Abstract
+
+In very large package registries, with a high frequency of releases, metadata overhead can lead to significant delays in common developer tasks. This is due to both the amount of metadata that needs to be downloaded, and the frequency with which it needs to be refreshed.
+
+Instead of scaling-up a single TUF repository, this TAP proposes scaling-out to multiple smaller repositories (eg. per-vendor, per-package, etc.) This dramatically reduces the rate at which metadata must be refreshed, since only packages actually used within a project are tracked.
+
+However, this approach introduces a Trust-On-First-Use (TOFU) issue for the root metadata for each of these TUF repos, as it would not be feasible to ship them with the client, as is currently specified. This challenge can be overcome with the use of a higher-level TUF repo, whose targets are the initial root metadata of the "sub-repos". Only the root metadata for the top-level TUF repository would have to ship with the client.
+
+The solution proposed in this TAP is only appropriate for loosely-coupled registries (see: [No Global Snapshot](#no-global-snapshot), below).
+
+# Motivation
+
+TUF provides several techniques to scale-up repositories and mitigate performance issues. However, these are proving insufficient for high-volume repositories that also have a high release frequency.
+
+## Hashed bin delegation
+
+[Hashed](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes) [bin](https://peps.python.org/pep-0458/#metadata-scalability) delegation is intended to improve performance by limiting the amount of metadata that clients must download. It uses a hashing algorithm to evenly distribute targets across the multiple metadata files (bins).
+
+While very significantly reduced, for large package registries, this still requires TUF clients to download a large volumes of metadata, even for projects using a relatively small number of packages.
+
+In large package registries that also have a high frequency of releases (eg. 1+/second), TUF clients will also have to regularly download updates to all of this bin metadata, *even if none of the project's dependencies have release a new version*.
+
+## Succinct hashed bin delegations (TAP-15)
+
+[Succinct hashed bin delegations](https://github.com/theupdateframework/taps/blob/master/tap15.md) reduces the size of `bins.json`, thus lowering the overall metadata that a client must download. However, this metadata *never* changes during normal operations and so only represents a relatively small metadata overhead.
+
+## Consistent snapshots
+
+[Consistent](https://theupdateframework.github.io/specification/latest/#consistent-snapshots) [snapshots](https://peps.python.org/pep-0458/#consistent-snapshots) are meant to address a specific challenge with volatile repositories: there is a high likelihood that a client tries to download a metadata file as it is being written. This would effectively result in a DOS.
+
+This strategy does not reduce the amount of metadata that must be downloaded by clients.
+
+## Snapshot Merkle Trees (TAP 16)
+
+[Snapshot Merkle Trees](https://github.com/theupdateframework/taps/blob/master/tap16.md) seeks to address the problem of Snapshot metadata growing in line with the number of hashed bins. Since Snapshot metadata must be downloaded with any update, this approach could reduce the metadata that clients must download by as much as 40%.
+
+However, it does not reduce amount of bin metadata, nor the churn inherent in frequent releases. This approach also increases complexity and requires a new third-party "auditor" role.
+
+
+# Rationale
+
+Very large, frequently updated package registries have large developer communities. Only a small minority of these developers every publish packages themselves. The vast majority will only even interact with the registry via a dependency manager. These users [?] are the target audience for TUF. As such, their primary use-cases (i.e. adding and updating packages within their projects) ought to be prioritized when considering TUF repository design (at the registry level).
+
+These use-cases involve frequent downloads ("read" operations) of TUF repository metadata. If these downloads negatively impact the developer experience (eg. quadrupling time for common client operations), we are likely to see resistance to adopting and using TUF. Therefore, minimizing the impact of these downloads on developers is the primary design goal for this TAP. 
+
+## High churn in TUF metadata using hashed bins
+
+Hashed bins bundle targets together, distributed evenly across multiple `bin_n` metadata files. A release of any package in the registry has an equal chance to update any given `bin_n` metadata. As a result, packages that are not in use for a given client project will still require re-download of `bin_n` metadata on an ongoing basis.
+
+Developers should expect to download new TUF metadata when there has been a release in a project dependency, since they will need to download a new version of the package anyway. However, using hashed bins, even if there have been no changes to any project packages, the client must re-download bin and snapshot metadata very frequently.
+
+To validate these hypotheses, the authors of this TAP created a [TAP-21 metadata overhead calculator](https://docs.google.com/spreadsheets/d/1Q1BPtS5T92e7Djx6878I0MdAOktg8hh73dEvMVhXIms) (based on the [calculator from PEP 458](https://docs.google.com/spreadsheets/d/11_XkeHrf4GdhMYVqpYWsug6JNz5ZK6HvvmDZX0__K2I)). The results for hashed bins (without this TAP) are outlined below:
+
+Assumptions:
+- (A1) Number of targets: 5M
+- (A2) Frequency of new releases: 60/minute
+- (A3) Number of bins: 16384
+- (A4) Number of dependencies in example project: 50
+- (A5) Average time between client updates: 60 minutes
+
+Calculations:
+- (C1) Likelihood that a bin will require an update in one minute: 0.37%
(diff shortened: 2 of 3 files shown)


JustinCappos (changes requested):
Can you remove the changes to tap1?   If they need to be part of a PR, they can be a separate one.

ergonlogic:
> Can you remove the changes to tap1? If they need to be part of a PR, they can be a separate one.

Yes, of course. I was separately experimenting with the idea of turning the headers into YAML front-matter. Not sure how those changes snuck in here.

tap1.md
ergonlogic, line 1: This change was included by mistake, and has been reverted.

tap21.md
```diff
@@ -0,0 +1,191 @@
...
+## Concerns
+
+### No global snapshot
+
+From the [TUF FAQ](https://theupdateframework.io/docs/faq/):
+> [T]he Snapshot role [...] provides a consistent view of the metadata available on a repo.
+
+One concern with the present proposal is that it lacks a global snapshot. See [Security Analysis](#security-analysis) (below) for a more detailed discussion on this topic.
+
+This concern is only relevant to registries where package versions are tightly-coupled (such as Debian Apt repositories). However, package versions are loosely-coupled in registries of language libraries (eg. Packagist, PyPi), which is the intended audience for this TAP.
```
Eh2406, line 129: Even a loosely coupled registry might want to provide consistent snapshots  to make mirroring the repository easier.  Setting up a mirror involves getting the latest snapshot, and downloading everything in the latest snapshot.  Updating a mirror involves getting the latest snapshot, diffing with the previous snapshot, and downloading the changes.  All of this is significantly harder without a snapshoting system.

ergonlogic: While snapshot metadata is convenient as a mirror manifest, presumably comparably diff'able data can be generated relatively easily from filesystem "modified" timestamps.

tap21.md
```diff
@@ -0,0 +1,195 @@
...
+## High churn in TUF metadata using hashed bins
+
+Hashed bins bundle targets together, distributed evenly across multiple `bin_n` metadata files. A release of any package in the registry has an equal chance to update any given `bin_n` metadata. As a result, packages that are not in use for a given client project will still require re-download of `bin_n` metadata on an ongoing basis.
+
+Developers should expect to download new TUF metadata when there has been a release in a project dependency, since they will need to download a new version of the package anyway. However, using hashed bins, even if there have been no changes to any project packages, the client must re-download bin and snapshot metadata very frequently.
+
+To validate these hypotheses, the authors of this TAP created a [TAP-21 metadata overhead calculator](https://docs.google.com/spreadsheets/d/1Q1BPtS5T92e7Djx6878I0MdAOktg8hh73dEvMVhXIms) (based on the [calculator from PEP 458](https://docs.google.com/spreadsheets/d/11_XkeHrf4GdhMYVqpYWsug6JNz5ZK6HvvmDZX0__K2I)). The results for hashed bins (without this TAP) are outlined below:
+
+Assumptions:
+- (A1) Number of targets: 5M
```
trishankatdatadog, line 67: Are these specifically all versions of all packages? One optimisation we did for PEP 458 is that we only sign the "indirect" simple HTML indices that point to the packages themselves. So instead of signing ~6M releases, we sign only ~600K simple indices.

tap21.md
```diff
@@ -0,0 +1,195 @@
...
+- (A2) Frequency of new releases: 60/minute
+- (A3) Number of bins: 16384
+- (A4) Number of dependencies in example project: 50
+
+Calculations:
+- (C1) Likelihood that a bin will require an update in one minute: 0.37%
+       [ C1 = 1 - ( (A3-1) / A3 )^A2 ]
+- (C2) Number of bins downloaded in example project: 100
```
trishankatdatadog, line 75: I think C2 can be reduced to A4 if the bins sign only the indices instead of the packages themselves.

tap21.md
```diff
@@ -0,0 +1,195 @@
...
+* TUF-Version: 1
+* Post-History: TBD
+
+# Abstract
+
+In very large package registries, with a high frequency of releases, metadata overhead can lead to significant delays in common developer tasks. This is due to both the amount of metadata that needs to be downloaded, and the frequency with which it needs to be refreshed.
+
+Instead of scaling-up a single TUF repository, this TAP proposes scaling-out to multiple smaller repositories (eg. per-vendor, per-package, etc.) This dramatically reduces the rate at which metadata must be refreshed, since only packages actually used within a project are tracked.
+
+However, this approach introduces a Trust-On-First-Use (TOFU) issue for the root metadata for each of these TUF repos, as it would not be feasible to ship them with the client, as is currently specified. This challenge can be overcome with the use of a higher-level TUF repo, whose targets are the initial root metadata of the "sub-repos". Only the root metadata for the top-level TUF repository would have to ship with the client.
```
trishankatdatadog, line 19: I like this idea of using a "high-level" TUF repo to distribute the root metadata for other "low-level" TUF "subrepos": it's something I had considered myself for other use cases (e.g., a company having multiple TUF repos for different packages, and wishing to have a cleaner separation between them, especially when there are no dependencies between these packages as appears to be your own use case).

However, I think that this is a different problem than what you are really describing below, and thus the solution doesn't fit your real problem.

trishankatdatadog (commented):
Thanks for sending your PR, Christopher and Derek! Left some comments based on one round of reading.

I think I see part of your problem but not its entire context. Do you think getting on a call might help?

tap21.md
```diff
@@ -0,0 +1,195 @@
...
+
+These use-cases involve frequent downloads ("read" operations) of TUF repository metadata. If these downloads negatively impact the developer experience (eg. quadrupling time for common client operations), we are likely to see resistance to adopting and using TUF. Therefore, minimizing the impact of these downloads on developers is the primary design goal for this TAP. 
+
+## High churn in TUF metadata using hashed bins
+
+Hashed bins bundle targets together, distributed evenly across multiple `bin_n` metadata files. A release of any package in the registry has an equal chance to update any given `bin_n` metadata. As a result, packages that are not in use for a given client project will still require re-download of `bin_n` metadata on an ongoing basis.
```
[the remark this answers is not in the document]
ergonlogic: TAP-21 does not suggest that clients need to download _all_ `bin_n` metadata.

Clients do need to download the latest version of each `bin_n` metadata that is relevant to the targets that they require. Each of those (relevant) `bin_n` metadata files include targets that are not required. New releases of _extraneous_ targets can produce new versions of _relevant_ `bin_n` metadata.

Therefore, clients need to download new `bin_n` metadata even if there have been no releases of required targets.

tap21.md
```diff
@@ -0,0 +1,195 @@
...
+## Snapshot Merkle Trees (TAP 16)
+
+[Snapshot Merkle Trees](https://github.com/theupdateframework/taps/blob/master/tap16.md) seeks to address the problem of Snapshot metadata growing in line with the number of hashed bins. Since Snapshot metadata must be downloaded with any update, this approach could reduce the metadata that clients must download by as much as 40%.
+
+However, it does not reduce amount of bin metadata, nor the churn inherent in frequent releases. This approach also increases complexity and requires a new third-party "auditor" role.
+
+
+# Rationale
+
+Very large, frequently updated package registries have large developer communities. Only a small minority of these developers every publish packages themselves. The vast majority will only even interact with the registry via a dependency manager. These developers are the target audience for TUF. As such, their primary use-cases (i.e. adding and updating packages within their projects) ought to be prioritized when considering TUF repository design (at the registry level).
```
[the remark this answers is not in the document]
ergonlogic: I'll update to "package managers".

I think in word-smithing that paragraph, we lost that explicit statement. I'll revise to make it clearer.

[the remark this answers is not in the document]
ergonlogic: See: 85c9037

tap21.md
```diff
@@ -0,0 +1,195 @@
...
+
+TUF provides several techniques to scale-up repositories and mitigate performance issues. However, these are proving insufficient for high-volume repositories that also have a high release frequency.
+
+## Hashed bin delegation
+
+[Hashed](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes) [bin](https://peps.python.org/pep-0458/#metadata-scalability) delegation is intended to improve performance by limiting the amount of metadata that clients must download. It uses a hashing algorithm to evenly distribute targets across the multiple metadata files (bins).
+
+While very significantly reduced, for large package registries, this still requires TUF clients to download a large volumes of metadata, even for projects using a relatively small number of packages.
+
+In large package registries that also have a high frequency of releases (eg. 1+/second), TUF clients will also have to regularly download updates to all of this bin metadata, *even if none of the project's dependencies have release a new version*.
```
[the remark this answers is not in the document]
ergonlogic: See my comment below: https://github.com/theupdateframework/taps/pull/189#discussion_r1866126794

tap21.md
ergonlogic: I agree that this solution seems promising for a number of other use-cases. However, we believe that it does directly address the challenges outlined in TAP-21's Motivation and Rationale sections.

> [...] I think that this is a different problem than what you are really describing below, and thus the solution doesn't fit your real problem.

Based on your earlier comments (eg. "I don't see [clients] needing to download all bin_n metadata, only what they need at the moment"), we may not have been sufficiently clear about the problem we're trying to solve. That said, if you think we've misidentified the root cause, I'm willing to re-evaluate.

tap21.md
trishankatdatadog: Yes: assumption is that the project indices contain at least the hashes, if not also the sizes. We can set a default upper bound on the file size in the TUF clients to avoid endless data attacks anyway.

tap21.md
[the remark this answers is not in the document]
trishankatdatadog: OK, I think I now see the issues (in increasing order of significance):
1. [The `bins` role metadata size](https://docs.google.com/spreadsheets/d/1Q1BPtS5T92e7Djx6878I0MdAOktg8hh73dEvMVhXIms/edit?gid=118020453#gid=118020453&range=B17) can be reduced quite a lot with [succinct hashed bin delegations](https://github.com/theupdateframework/taps/blob/master/tap15.md).
2. Do you _need_ hashes (SHA2-256 _and_ SHA2-512) in your snapshot metadata? You really only need it for protection against malicious mirrors (see Section 5.6 [here](https://www.usenix.org/system/files/conference/atc17/atc17-kuppusamy.pdf)). If not, you should be to shave a lot off your snapshot by distributing only version numbers. Even if you do need hashes, have you calculated what [snapshot Merkle trees](https://github.com/theupdateframework/taps/blob/master/tap16.md) can buy you?
3. What are the dependencies between your packages? How do you plan to solve for mix-and-match and rollback attacks there?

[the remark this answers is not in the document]
ergonlogic: > 1. The bins role metadata size can be reduced quite a lot with succinct hashed bin delegations.

We address this in the Motivation section: [Hashed bin delegation](https://github.com/consensus-enterprises/taps/blob/scalability-tap/tap21.md#succinct-hashed-bin-delegations-tap-15). TAP-21 is primarily concerned with the churn of `bin_n` metadata. `bins` metadata, by virtue of being essentially static, does not factor into it.

Also, since we expect the top-level repo to require hashed bins, implementing TAP-15 would reduce overall TUF metadata when implementing TAP-21 as well.

> 2. Do you need hashes (SHA2-256 and SHA2-512) in your snapshot metadata? 

Including both SHA2-256 and SHA2-512 does seem redundant. I've updated the calculator spreadsheet to reflect removing SHA2-512 hashes. This streamlining also reduces both top-level and sub-repo metadata sizes (though the latter is not reflected in the calculations yet).

> You really only need it for protection against malicious mirrors (see Section 5.6 here). If not, you should be to shave a lot off your snapshot by distributing only version numbers.

Very large repositories are almost certainly going to use mirrors. In my (admittedly limited) experience, they do want to protect against the potential of malicious mirrors. I proposed just distributing version numbers, and was asked to keep the hashes.

Hoever, this streamlining would also reduce both top-level and sub-repo metadata sizes in TAP-21.

> Even if you do need hashes, have you calculated what snapshot Merkle trees can buy you?

Not directly, no. We did address this in the Motivation section: [Snapshot Merkle Trees (TAP 16)](https://github.com/consensus-enterprises/taps/blob/scalability-tap/tap21.md#snapshot-merkle-trees-tap-16). For ease of comparison, I assumed a 90% reduction in snaphot metadata size using TAP-16. While reducing overall download size, this has no effect on `bin_n` churn.

Also, a [recent Slack discussion on TAP-16](https://cloud-native.slack.com/archives/C8NMD3QJ3/p1732032870278729) suggests there may be other performance issues with this approach.

> 3. What are the dependencies between your packages?

For our primary use case, package dependencies are explicitely listed in the registry indexes, and resolved (independently of TUF) by the package manager. These are what we refer in the TAP as "loosely-coupled" in the TAP.

> How do you plan to solve for mix-and-match and rollback attacks there?    

We address both [mix-and-match](https://github.com/consensus-enterprises/taps/blob/scalability-tap/tap21.md#mix-and-match-attack) and [rollback](https://github.com/consensus-enterprises/taps/blob/scalability-tap/tap21.md#roll-back-attack) attacks in the Security Analysis section. Did you have a specific concern that we did not address?

tap21.md
ergonlogic: > Are these specifically all versions of all packages? One optimisation we did for PEP 458 is that we only sign the "indirect" simple HTML indices that point to the packages themselves. So instead of signing ~6M releases, we sign only ~600K simple indices.

As noted elsewhere in this PR, this reduces the number of `bin_n` metadata required for a project to just those containing the required indices; as opposed to the bins for both indices and packages. This reduces the overall size of the downloaded metadata per update.

However, the frequency at which TUF metadata changes remains constant, since each new package release requires an update to the associated index. Thus, it does not reduce the frequency that `bin_n` metadata requires an update. Both the dividend and the divisor are halved in the [relevant calculations](https://github.com/consensus-enterprises/taps/blob/scalability-tap/tap21.md#high-churn-in-tuf-metadata-using-hashed-bins):

> (C2) Number of bins downloaded in example project: 100 [ C2 = 2 x A4 ] (1 for each of package and index)
> (C3) Average number of bins requiring update per minute: 0.3656 [ C3 = C1 x C2 ]
> (C4) Time to require full metadata refresh: 274 minutes [ C4 = C2/C3 ]

Drastically reducing the number of targets suggests that the number of bins should also be reduced. This, perhaps counterintuitively, results in higher `bin_n` churn:

```
Assumptions:

    (A1) Number of targets: 600K

    (A2) Frequency of new releases: 60/minute

    (A3) Number of bins: 2048

    (A4) Number of dependencies in example project: 50

Calculations:

    (C1) Likelihood that a bin will require an update in one minute: 2.89% [ C1 = 1 - ( (A3-1) / A3 )^A2 ]

    (C2) Number of bins downloaded in example project: 50 [ C2 = A4 ] (only indices)

    (C3) Average number of bins requiring update per minute: 1.4439

    [ C3 = C1 x C2 ]

    (C4) Time to require full metadata refresh: 35 minutes [ C4 = C2/C3 ]

```

Note that A1 is not used in any of these calculations. It merely informs the [optimal number of bins](https://gist.github.com/petres/8698f213f11e86c63cd9d83921a3a166).

tap21.md
trishankatdatadog: Let's continue discussing in our video meetings, and we'll get back here once we agree on some ways forward.

aparcar:
Looks like TAP-21 is now already taken bei ML-DSA, should this be renamed? #195

JustinCappos:
> Looks like TAP-21 is now already taken bei ML-DSA, should this be renamed? #195

Yes, the actual TAP number usually only gets assigned when it is officially accepted as a draft.  So, removing the number for now is the best outcome.  When it's ready to move to draft status, then the TAP editors will review it and after approval, a number can be assigned.

cyphar:
As I mentioned to @JustinCappos at KubeCon, at @amutable-systems we have been trying to solve a different problem by splitting out repos and creating cross-repo links, but unfortunately the approach here is not really flexible enough for us.

I will publish a blog post about this in a week or two but I'm just leaving a comment here now as a reminder (I only managed to find this PR by chance!).

PR #195, opened 2026-05-04
ML-DSA signing scheme for TUF metadata

kommendorkapten:
Wrote up a TAP on how to encode ML-DSA keys when used within TUF and added a section on pre-auth encoding as certain targets can be fairly larger, and ML-DSA operates in pure mode, that is, no pre-hashing which is the case for RSA and ECDSA.

### DIFF
--- a/tap21.md
+++ b/tap21.md
@@ -1,0 +1,181 @@
+* TAP: 21
+* Title: ML-DSA signing scheme for TUF metadata
+* Last-Modified: 2026-04-30
+* Author: Fredrik Skogman
+* Status: Draft
+* Content-Type: text/markdown
+* Created: 2026-04-29
+
+# Abstract
+
+This TAP proposes an application-level pre-hashing scheme to use with
+ML-DSA that minimizes the data sent to the signing device. Instead of
+passing the full canonicalized metadata, the application computes a
+cryptographic hash over the metadata and computes a pre-signing byte
+string (including domain separator and protocol version). The
+pre-signing byte string is what is sent to the signing device. This
+approach keeps the HSM interface simple and bounded while preserving
+the security properties required by FIPS 204.
+
+# Motivation
+
+TUF metadata can be large, particularly for targets metadata in
+repositories with many artifacts. ML-DSA pure mode signing requires
+the entire message to be available to the signing device. When private
+keys are held in hardware security modules (HSMs), the HSM must
+receive the full message to produce a pure mode signature.
+Transmitting large metadata payloads to an HSM introduces practical
+limitations on message size, and potential interface constraints that
+make pure mode ML-DSA unsuitable for some TUF deployments.
+
+# Specification
+
+Conventions used:
+
+* `0x__`: a raw byte value, specified as a hexadecimal number
+* `||`: byte concatenation
+
+The raw TUF metadata is NEVER signed, instead a pre-signing byte string is
+created with the following format, offering domain separation and
+versioning at the protocol level:
+
+```
+0x74 || 0x75 || 0x66 || <version> || H(MSG)
+```
+
+The domain separators are the ASCII codes for `tuf`.
+
+Version must be a single byte specifying the version. `0x01` is
+currently the only defined version.
+
+The hash function and the canonicalization scheme for the message are
+specified by the version.
+
+Pure ML-DSA MUST be used with an **empty context**.
+
+## Rationale
+
+Why not use the `scheme` to specify the hash algorithm and instead use
+the version? The version specifies the entire set of choices and
+cryptographically binds those choices to the signature which would
+reduce risks of misuse. This allows for easier updates of the
+versioning scheme as it's an all-or-nothing approach. By providing
+some details via the scheme and others via some versioning opens up
+for possible confusion.
+
+Why not use HashML-DSA? With Ed25519 the ecosystem support has been
+much better for the pure version, and it's likely it will be the same
+for ML-DSA. Based on this an application specific protocol is better
+suited for wider adoption.
+
+## Protocol versions
+
+To allow for future updates on hash algorithm selection to mitigate
+any collision or preimage attack the selection of hash algorithm is
+specified via a protocol version. This provides a layer of indirection
+where certain details can change over time without encoding too much
+information into the `scheme` parameter.
+
+### v1
+
+* Version byte: `0x01`
+* Hash algorithm: SHA-512
+* Implementations MUST support
+    * `ML-DSA-44`
+    * `ML-DSA-65`
+    * `ML-DSA-87`
+* Metadata canonicalization scheme: encoded as "canonical JSON" as described
+  in the [TUF
+  Specification](https://theupdateframework.github.io/specification/v1.0.34/index.html#metaformat).
+
+## Signature generation
+
+1. Load the public key from TUF metadata
+2. Parse the version from the public key's `scheme` and prepare the
+   hash function `H`
+3. Compute the canonicalized metadata representation `MSG`
+4. Create the pre-signing byte string:
+    ```
+    0x74 || 0x75 || 0x66 || version || H(MSG)
+    ```
+5. Sign the pre-signing byte string using an empty context
+
+## Verification steps:
+
+1. Compute canonical metadata representation
+2. Load up the public key for verification
+3. Parse `scheme` into parameter set and version
+    * Reject if the protocol version is not supported
+    * Implementations MUST NOT infer or select an ML-DSA
+      parameter set or version from the signature bytes alone
+4. Verifier must reconstruct the exact signed bytes itself
+    * It should not accept a caller-supplied digest/prefix blindly
+    * The `version` MUST be taken from the trusted TUF metadata's
+      `scheme` parameter
+    * The version binds hash function to use
+    * It should compute: `digest = H(MSG)`
+    * Then verify over `0x74 || 0x75 || 0x66 || version || digest`
+      using an empty context
+5. Reject unknown or mismatched versions
+    * Do not fall back
+    * Do not try multiple interpretations
+    * Do not accept the same signature under HashML-DSA or another scheme
+
+## TUF metadata parameters:
+
+* `keytype`: `ml-dsa`
+* `scheme`: (`ml-dsa-<parameter set>/<version>`)
+    * `ml-dsa-44/1`
+    * `ml-dsa-65/1`
+    * `ml-dsa-87/1`
+* `keyval.public`: PEM encoding of DER-encoded `SubjectPublicKeyInfo`
+  structures as defined for ML-DSA in RFC 9881
+* `signature.sig`: Hex-encoded signature byte string as per FIPS 204
+   §7.2
+
+# Security considerations
+
+1. SHA-512 length extension: Not a concern. The signature is made over `domain
+   || version || digest`. Length extension is more a concern when the
+   digest is computed to be used as a MAC
+2. SHA-512 vs ML-DSA-87 margin. SHA-512 has zero margin, but is
+   valid. Future versions can increase the margin if deemed necessary
+3. Signature confusion/replay: With the domain separation a valid
+   signature over the raw digest from another domain would not be
+   valid in the TUF metadata domain (collisions on the domain and
+   version bytes are _very_ unlikely)
+
+# Appendix
+
+## ML-DSA parameter sets
+
+This is an excerpt only, see FIPS-204 §4 for a complete parameter set.
+
+| Parameter | ML-DSA-44 | ML-DSA-65 | ML-DSA-87 |
+|-----------|-----------|-----------|-----------|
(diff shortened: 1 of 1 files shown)


tap21.md
```diff
@@ -0,0 +1,181 @@
...
+TUF metadata can be large, particularly for targets metadata in
+repositories with many artifacts. ML-DSA pure mode signing requires
+the entire message to be available to the signing device. When private
+keys are held in hardware security modules (HSMs), the HSM must
+receive the full message to produce a pure mode signature.
+Transmitting large metadata payloads to an HSM introduces practical
+limitations on message size, and potential interface constraints that
+make pure mode ML-DSA unsuitable for some TUF deployments.
+
+# Specification
```
mnm678, line 31: I was a bit confused until I got to the "TUF metadata parameters" section about where this fit with TUF. Maybe you could move that sub-section earlier

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. Load the public key from TUF metadata
+2. Parse the version from the public key's `scheme` and prepare the
+   hash function `H`
+3. Compute the canonicalized metadata representation `MSG`
+4. Create the pre-signing byte string:
+    ```
+    0x74 || 0x75 || 0x66 || version || H(MSG)
+    ```
+5. Sign the pre-signing byte string using an empty context
```
mnm678, line 128: Does it make sense for us to include advice around where to do the signing? (Like do we suggest using a hardware device?) I'm not sure if this makes sense to include, but maybe we can include a link or something to guide implementers in the right direction

mnm678 (dismissed):
I left a few minor comments, but overall I think this is a great scheme, and addresses some of the performance concerns around ML-DSA

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+
+The raw TUF metadata is NEVER signed, instead a pre-signing byte string is
+created with the following format, offering domain separation and
+versioning at the protocol level:
+
+```
+0x74 || 0x75 || 0x66 || <version> || H(MSG)
+```
+
+The domain separators are the ASCII codes for `tuf`.
```
Hayden-IO, line 46: love this :)

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+
+Why not use the `scheme` to specify the hash algorithm and instead use
+the version? The version specifies the entire set of choices and
+cryptographically binds those choices to the signature which would
+reduce risks of misuse. This allows for easier updates of the
+versioning scheme as it's an all-or-nothing approach. By providing
+some details via the scheme and others via some versioning opens up
+for possible confusion.
+
+Why not use HashML-DSA? With Ed25519 the ecosystem support has been
```
Hayden-IO, line 83: There's the practical lack of implementation, and there's also https://keymaterial.net/2024/11/05/hashml-dsa-considered-harmful/

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+    * `ML-DSA-65`
+    * `ML-DSA-87`
+* Metadata canonicalization scheme: encoded as "canonical JSON" as described
+  in the [TUF
+  Specification](https://theupdateframework.github.io/specification/v1.0.34/index.html#metaformat).
+
+## Signature generation
+
+1. Load the public key from TUF metadata
+2. Parse the version from the public key's `scheme` and prepare the
```
Hayden-IO, line 121: We could change `H` to `SHA-512` since it's fixed? I think this was from an earlier revision when `H` was variable.

tap21.md
Hayden-IO: I think this should be out of scope for the TAP, as the decision to use an HSM vs SK vs KMS vs raw key is based on the integrating project's threat model.

Maybe as a separate TAP if this guidance doesn't exist?

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+specified via a protocol version. This provides a layer of indirection
+where certain details can change over time without encoding too much
+information into the `scheme` parameter.
+
+### v1
+
+* Version byte: `0x01`
+* Hash algorithm: SHA-512
+* Implementations MUST support
+    * `ML-DSA-44`
```
Hayden-IO: Should we explicitly state the valid `KEYTYPE`s and `SCHEME`s? All lower case to match https://theupdateframework.github.io/specification/latest/#file-formats-keys or upper case?

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+    ```
+5. Sign the pre-signing byte string using an empty context
+
+## Verification steps:
+
+1. Compute canonical metadata representation
+2. Load up the public key for verification
+3. Parse `scheme` into parameter set and version
+    * Reject if the protocol version is not supported
+    * Implementations MUST NOT infer or select an ML-DSA
```
Hayden-IO, line 136: Should implementations verify the length of the public key and/or signature based on the ML-DSA parameter set? Maybe this is already handled by the underlying library?

tap21.md
Hayden-IO: ah you do below...maybe that should be higher?

tap21.md
```diff
@@ -0,0 +1,181 @@
...
+
+# Appendix
+
+## ML-DSA parameter sets
+
+This is an excerpt only, see FIPS-204 §4 for a complete parameter set.
+
+| Parameter | ML-DSA-44 | ML-DSA-65 | ML-DSA-87 |
+|-----------|-----------|-----------|-----------|
+| 𝜆         | 128       | 192       | 256       |
```
Hayden-IO: what's the purpose of including this?

Hayden-IO (commented):
This is fantastic!

tap21.md
kommendorkapten: This is a remnant from when the hash algorithm was bound to the parameter set, to provide details on the security level required for each. I'll remove this as it's not relevant.

tap21.md
kommendorkapten: Yes, @mnm678 had a few similar comments to I'll shuffle things around so more TUF details are in the beginning, I think that will make it easier to read.

tap21.md
kommendorkapten: I think we need to sit tight here on any recommendation. I immediately see two different paths in the future:
* Root signing like Sigstore does: HSMs, like YubiKey (assuming they release one soon that supports ML-DSA)
* KMS for online signing of eg. targets. 

So I think this TAP benefits from just defining the signing scheme.

tap21.md
kommendorkapten: I added a link to that too, thanks.

tap21.md
kommendorkapten: My thinking of keeping it is that we still have the hash decided per the version of the protocol. But you are right that for `v1` only SHA-512 is allowed. For keeping the spec generic of the version I think it makes sense to keep it as is.

tap21.md
kommendorkapten: That *should* be managed by the underlying library, and I rather not see any TUF related code try to delve into analyzing the signature/pk size. I clarify this.

kommendorkapten:
Thanks for the feedback ❤️ 
All comments should be adressed/answered. See https://github.com/kommendorkapten/taps/blob/pqc/tap21.md for a rendered version.

Hayden-IO (approved):
Really great work on this spec!

mnm678 (dismissed):

tap21.md
```diff
@@ -0,0 +1,190 @@
...
+
+The domain separators are the ASCII codes for `tuf`.
+
+Version must be a single byte specifying the version. `0x01` is
+currently the only defined version.
+
+The hash function and the canonicalization scheme for the message are
+specified by the version.
+
+Pure ML-DSA MUST be used with an **empty context**.
```
jku, line 54: Out of curiosity,  what is the reason for not using `context=0x74 || 0x75 || 0x66 || version` but instead building the whole pre-signing string ourselves?

jku (commented):
Looks good to me. Left one question

jku (approved):

tap21.md
jku: is this what your comment about "μ" is about? I could not follow that one.

kommendorkapten: The ML-DSA Sign function is roughly defined as `Sign(private_key, msg, context)` where it internally computes a hash μ and then calls `SignInternal(private_key, μ)` (this is a gross over-simplification though). So this means that a compliant ML-DSA can split the internal pre-hashing so that e.g. an HSM can compute the hash client side and only send a smaller payload over the wire. OpenSSL 4.0 exposes this API so clients can use this functionality directly. OpenSSL calls this "ML-DSA-MU". It's not unlikely other crypto libs will expose similar APIs so that's why I added it to the rationale on why we are _not_ using that approach.

Putting the domain separation in the `context` would yield the same result, but we would still have to define the pre-signing schema (hashing and canonical message representation) so I figure it's just easier to make it into a single input over spreading it onto two fields.

jku: Got it. Initially it seemed like our pre-signing schema was weaker (with plain concatenation of the parts without limiters) but that actually won't matter since our data is a hash: no chance of the context string and data getting mixed up. LGTM

jku:
For testing: https://jku.github.io/pqc-test-repo/metadata/ is a live repository with ML-DSA-44 online keys that should match this proposal

JustinCappos:
> with ML-DSA-44 online keys 

Note for anyone else who is glancing quickly at this early in the AM, only the snapshot / timestamp keys are ML-DSA-44, as I should have expected from the use of the word online.   😆 

We probably should list some ballpark numbers in the TAP for the metadata size (key / sig size differences) and signing / verification time differences.

kommendorkapten:
Yeah, I can add that.

jku:
## TUF Metadata Size Analysis: ECDSA to ML-DSA Transition

This uses the sigstore repository as an example (5 root/targets keys, 1 timestamp/snapshot key).

Note that I did not actually create a working repository like this: this result comes from a script that calculates the sizes, script is made by an AI -- I've made point verifications  but feel free to do you own homework. 

| TUF Role  | Keys | Sigs | ECDSA Size | ML-DSA-44 Size (% Increase) | ML-DSA-65 Size (% Increase) | ML-DSA-87 Size (% Increase) |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| root | 6 | 5 | 5,630 B | **39,326 B** (+598.5%) | **53,406 B** (+848.6%) | **71,968 B** (+1,178.3%) |
| targets | 1 | 5 | 4,942 B | **30,133 B** (+509.7%) | **39,813 B** (+705.6%) | **53,965 B** (+992.0%) |
| timestamp | 0 | 1 | 447 B | **5,147 B** (+1,051.5%) | **6,907 B** (+1,445.2%) | **9,561 B** (+2,038.9%) |
| snapshot | 0 | 1 | 1,760 B | **6,458 B** (+266.9%) | **8,218 B** (+366.9%) | **10,872 B** (+517.7%) |
| registry.npmjs.org | 0 | 1 | 642 B | **5,338 B** (+731.5%) | **7,098 B** (+1,005.6%) | **9,752 B** (+1,419.0%) |

For this specific case the timestamp growth is the most worrying: that is the only file that every single client downloads constantly. 

The table is based on these estimates:

**Public keys:**

  - **ECDSA**: DER size= **91 bytes**. PEM size = **178 bytes**. On-disk size inside JSON literal = **180 bytes**.
  - **ML-DSA-44**: DER size= **1334 bytes**. PEM size = **1,860 bytes**. On-disk size inside JSON literal = **1,888 bytes**.
  - **ML-DSA-65**: DER size= **1974 bytes**. PEM size = **2,726 bytes**. On-disk size inside JSON literal = **2,768 bytes**.
  - **ML-DSA-87**: DER size= **2614 bytes**. PEM size = **3,595 bytes**. On-disk size inside JSON literal = **3,650 bytes**.

**signatures**

  - **ECDSA**:_ **140 hex characters**
  - **ML-DSA-44**: **4,840 hex characters**
  - **ML-DSA-65**: **6,600 hex characters**
  - **ML-DSA-87**: **9,254 hex characters**

kommendorkapten:
Updated with three summary tables on the expected growth from ECDSA inspired by @jku's comment.

JustinCappos (approved):
As discussed in the TUF community meeting today, we think this is in sufficient state to merge as a draft / assign TAP number 21 to this.

Hayden-IO:
Just wanted to check in on the status of getting this merged.

JustinCappos:
I'd also like to point out that we are looking for TAP editors.
https://github.com/theupdateframework/taps/issues/197  If dont' know if
this is blocking on waiting for review there, but more eyeballs to review
may help.  (Review load is very minimal)

mnm678 (approved):

mnm678:
This is ready to merge as a draft. Once https://github.com/theupdateframework/go-tuf/pull/758 moves forward we can also look at getting this TAP accepted.

PR #196, opened 2026-05-29
Document TAP 11 on POUFs for TUF interoperability

azariahporras-blip:
This TAP describes the POUF mechanism for standardizing interoperability in TUF implementations, detailing its structure, motivation, and security considerations.


JustinCappos (changes requested):
PR removes the TAP text.  Looks like either an AI bot gone wrong or someone griefing us.

